PR #152, opened 2020-08-11
feat: add leasing interest call

Averethel:
Adds `sendLeasingInterest` api call. It will behave the same way message leas call behaves in terms of validation.

~~**note** the validation part is not yet implemented in the backend~~

### DIFF
--- a/src/services/car/__tests__/leasing.test.ts
+++ b/src/services/car/__tests__/leasing.test.ts
@@ -1,0 +1,25 @@
+import { fetchLeasingFormUrl } from "../leasing"
+
+describe("CAR service", () => {
+  beforeEach(fetchMock.resetMocks)
+
+  describe("fetchLeasingFormUrl", () => {
+    it("returns the url", async () => {
+      fetchMock.mockResponse(JSON.stringify({ url: "test-url" }))
+
+      expect(
+        await fetchLeasingFormUrl({ listingId: 123, locale: "de" })
+      ).toEqual("test-url")
+    })
+
+    it("returns null if listing fails validation rules", async () => {
+      fetchMock.mockResponse(JSON.stringify({ message: "too expensive" }), {
+        status: 422,
+      })
+
+      expect(
+        await fetchLeasingFormUrl({ listingId: 123, locale: "de" })
+      ).toEqual(null)
+    })
+  })
+})
--- a/src/services/car/__tests__/leasing.test.ts
+++ b/src/services/car/__tests__/leasing.test.ts
@@ -1,4 +1,4 @@
-import { fetchLeasingFormUrl } from "../leasing"
+import { fetchLeasingFormUrl, sendLeasingInterest } from "../leasing"
 
 describe("CAR service", () => {
   beforeEach(fetchMock.resetMocks)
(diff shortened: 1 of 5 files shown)


src/services/car/leasing.ts
```diff
@@ -35,3 +36,35 @@ export const fetchLeasingFormUrl = async (args: {
       return null
   }
 }
+
+export const sendLeasingInterest = async (
```
lkappeler, line 40: Can wen envelope the arguments for me this makes it always easier to understand the interface...
Same as for `fetchLeasingFormUrl`

src/services/car/__tests__/leasing.test.ts
```diff
@@ -22,4 +22,117 @@ describe("CAR service", () => {
...
+      })
+
+      describe("validate only", () => {
+        it("calls validation endpoint", async () => {
+          await sendLeasingInterest(12345, leasingInterest(), {
+            validateOnly: true,
+          })
+
+          expect(fetch).toHaveBeenCalledWith(
+            expect.stringMatching(
```
lkappeler, line 47: Whats the reason to use a regex over the exact path as string?

src/services/car/__tests__/leasing.test.ts
```diff
@@ -22,4 +22,117 @@ describe("CAR service", () => {
...
+      })
+
+      it("returns a failure", async () => {
+        const result = await sendLeasingInterest(
+          12345,
+          leasingInterest({ email: "test@test" })
+        )
+
+        expect(result.tag).toEqual("error")
+        if (result.tag === "error") {
```
lkappeler, line 114: The condition does not make a lot of sense if the tag is not error the expect above will throw...

src/services/car/leasing.ts
```diff
@@ -35,3 +36,35 @@ export const fetchLeasingFormUrl = async (args: {
...
+  listingId: number,
+  leasingInterest: LeasingInterest,
+  options = {}
+): Promise<WithValidationError<LeasingInterest>> => {
+  const { validateOnly, recaptchaToken } = {
+    validateOnly: false,
+    recaptchaToken: null,
+    ...options,
```
lkappeler, line 48: What are you spreading the options for here if you only destructure the defined values anyways?

src/services/car/__tests__/leasing.test.ts
```diff
@@ -22,4 +22,117 @@ describe("CAR service", () => {
...
+          expect(result.errors).toEqual(errors)
```
lkappeler, line 115: Isn't this testing weather the mock gets applied to the mock response?

src/services/car/__tests__/leasing.test.ts
```diff
@@ -22,4 +22,117 @@ describe("CAR service", () => {
...
+          status: 500,
+        })
+      })
+
+      it("returns a failure", async () => {
+        const result = await sendLeasingInterest(12345, leasingInterest())
+
+        expect(result.tag).toEqual("error")
+        if (result.tag === "error") {
+          expect(result.message).toEqual("validation.other-error")
```
lkappeler, line 133: Shouldn't this be covered by the unit test of the `handleValidationError` function?

src/types/models/index.ts
```diff
@@ -129,6 +129,14 @@ export interface MessageLead {
   }
 }
 
+export interface LeasingInterest {
```
lkappeler, line 132: We can send a leasing interest without any information about the customer?

src/services/car/leasing.ts
Averethel: I wanted to keep it consistent with `sendMessageLead` since it'll be used in a similar context. We have a mix of both in the project let's tackle unifying the style within a tech-20 https://autoricardo.atlassian.net/browse/CAR-4826

src/services/car/__tests__/leasing.test.ts
Averethel: `fetch` takes a full URL as an argument. I'd rather not parse it but just match the path

src/services/car/__tests__/leasing.test.ts
Averethel: Indeed. However, to be able to access properties of the `error` tagged response you need to ensure that the `tag` is indeed `error`. Long story short jests expect and an exception it throws is not enough for typescript to know you're in that branch.

src/services/car/leasing.ts
Averethel: Those are default values we want to use if nothing is provided.
If we have:
```javascript
const obj = { foo: "bar" }

const { foo } = { 
  foo: "default",
  ...obj,
}
```
`foo` would equal `"bar"` since the spread from `obj` overrides default.

src/services/car/__tests__/leasing.test.ts
Averethel: It's testing that the error is properly handled, no?

src/services/car/leasing.ts
lkappeler: Mhh ofc misread that, thanks for the explanation

src/services/car/__tests__/leasing.test.ts
Averethel: Oh, I get what you mean above now. I guess we could test that we call `handleValidationError` instead, but I was never a big fan of testing whether the error handling method was called, but would rather test the result. WDYT?

src/types/models/index.ts
Averethel: We can send it to the backend and it would not pass validation. We have optional values to support the empty form.

Averethel:
@GutierrezGo feel free to use the pre-release for the modal form and merge this when you're ready

src/services/car/__tests__/leasing.test.ts
lkappeler: Yeah it's fine just for my understanding, no strong preference :)

src/types/models/index.ts
lkappeler: Ahh makes sense

PR #155, opened 2020-08-20
feat: add dealer promotion apis

ufkaya:
BREAKING CHANGE

- [CAR-4653 - Garage Carousel: Edit Page - Slogan, Date and Photo Upload](https://autoricardo.atlassian.net/browse/CAR-4653)

- [x] New APIs to get, create and modify the promotion
- [x] New API to buy a product by defining the start date

### DIFF
--- a/src/services/product.ts
+++ b/src/services/product.ts
@@ -9,7 +9,7 @@ export const fetchProducts = async (): Promise<Product[]> => {
   return fetchPath(Service.DEALER, "products")
 }
 
-export const purchaseAndUseListingProduct = async (
+export const purchaseAndUseProduct = async (
   dealerId: number,
   listingId: number,
   productId: number
@@ -33,7 +33,7 @@ export const purchaseAndUseListingProduct = async (
   })
 }
 
-export const purchaseAndUseProduct = async (
+export const purchaseAndUseProductWithStartDate = async (
   dealerId: number,
   productId: number,
   startDate: string
--- a/src/services/product.ts
+++ b/src/services/product.ts
@@ -9,7 +9,7 @@ export const fetchProducts = async (): Promise<Product[]> => {
   return fetchPath(Service.DEALER, "products")
 }
 
-export const purchaseAndUseProduct = async (
+export const purchaseAndUseListingProduct = async (
   dealerId: number,
   listingId: number,
   productId: number
@@ -33,7 +33,7 @@ export const purchaseAndUseProduct = async (
   })
 }
 
-export const purchaseAndUseProductWithStartDate = async (
+export const purchaseAndUseDealerProduct = async (
   dealerId: number,
   productId: number,
   startDate: string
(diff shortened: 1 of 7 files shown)


ralucacril (approved):

src/services/product.ts
```diff
@@ -32,3 +32,28 @@ export const purchaseAndUseProduct = async (
     }
   })
 }
+
+export const purchaseAndUseProductWithStartDate = async (
```
Averethel, line 36: Can we go with different naming? I thought that this is the same product but with an additional `startDate` parameter at a first glance.

How about we:
1. rename `purchaseAndUseProduct` to `purchaseAndUseListingProduct`
2. call this `purchaseAndUseGlobalProduct` or `purchaseAndUseDealerProduct`?

WDYT?

src/types/models/dealerPromotion.ts
```diff
@@ -0,0 +1,13 @@
+interface Data {
```
Averethel, line 1: How about:

```suggestion
interface PromotionContent {
```

src/types/models/dealerPromotion.ts
```diff
@@ -0,0 +1,13 @@
...
+  description?: string
+  title: string
+}
+
+export interface DealerPromotion {
+  dataDe?: Data
```
Averethel, line 7: Is this naming enforced by BE? Can we go with something less generic?

Averethel (changes requested):
A few suggestions from my side. Other than naming in the `product` service everything is optional.
Please also link new services/calls in the readme

src/services/product.ts
ufkaya: ok, it is similar to my first approach -> https://github.com/carforyou/carforyou-api-client-pkg/pull/155/commits/3a3af3c31d1352e476f33798e8f500e85b887823

I will rename them to `purchaseAndUseListingProduct` and `purchaseAndUseDealerProduct`

src/types/models/dealerPromotion.ts
ufkaya: Yes, https://dealer-service.preprod.carforyou.ch/swagger-ui/index.html#/Dealer/createPromotionUsingPOST

Averethel: Can you check with them? It should be a low effort to rename a database field and if it makes our code more readable it's definitely worth it.

ufkaya: And how would you rename them?

masone: maybe

```
{ data: {
  de: {},
  fr: {},
 ...
}
```

but I'd do whatever backend is comfortable with. would not change the database if we follow this naming pattern in other places already

ufkaya: @Averethel I checked with BE, is not possible, they implemented it this way e2e

lkappeler:
@ufkaya Can you add the new endpoints to the readme, thanks

Averethel (approved):
Looks good to me. Please link the new calls in the readme.
Don't forget that this is a `BREAKING CHANGE` since we renamed the purchase product call

PR #161, opened 2020-08-26
feat: add type definitions for the dealer promotion feature

lkappeler:
Adds and updates some types for the dealer promotion feature

### DIFF
--- a/src/types/models/dealerPromotion.ts
+++ b/src/types/models/dealerPromotion.ts
@@ -3,6 +3,11 @@ interface PromotionContent {
   title: string
 }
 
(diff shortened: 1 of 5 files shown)


src/types/models/dealerPromotion.ts
```diff
@@ -3,6 +3,11 @@ interface PromotionContent {
   title: string
 }
 
+export interface DealerPromotionContent extends PromotionContent {
```
ufkaya, line 6: Can you extend the `DealerPromotion` with this one?

ufkaya (approved):
👍

PR #164, opened 2020-09-02
feat: accept language parameter in dealer fetch

Averethel:
Due to dealer promotion being created in multiple languages the `fetchDealer` call accepts an optional `language` option to get the right language version of promotion content.

### DIFF
--- a/src/services/dealer.ts
+++ b/src/services/dealer.ts
@@ -19,12 +19,8 @@ export const fetchDealer = async (
   id: number,
   options: { language?: "de" | "en" | "fr" | "it" } = {}
 ): Promise<Dealer> => {
-  return fetchPath(
-    Service.DEALER,
-    `dealers/${id}${
-      Object.keys(options).length ? `?${toQueryString(options)}` : ""
-    }`
-  )
+  const query = toQueryString(options)
+  return fetchPath(Service.DEALER, `dealers/${id}${query ? `?${query}` : ""}`)
 }
 
 export const fetchDealerSuggestions = async (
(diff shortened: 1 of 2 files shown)


src/services/dealer.ts
```diff
@@ -13,8 +13,16 @@ import { DealerPromotion } from "../types/models/dealerPromotion"
...
-  return fetchPath(Service.DEALER, `dealers/${id}`)
+import toQueryString from "../lib/toQueryString"
+
+export const fetchDealer = async (
+  id: number,
+  options: { language?: "de" | "en" | "fr" | "it" } = {}
+): Promise<Dealer> => {
+  return fetchPath(
+    Service.DEALER,
+    `dealers/${id}${options.language ? `?${toQueryString(options)}` : ""}`
```
masone, line 24: sooo if we pass a language, we convert all options to a query string? this will result in other fields than just the language being included

1. if you want to be explicit about passing the options, only check and pass the language option
2. if you want to be flexible, convert all options to query string and append it always when there is any option

masone (changes requested):

src/services/dealer.ts
ufkaya, line 24: Can toQueryString handle empty objects? If yes, you can remove `options.language ?`

ufkaya (approved):

src/services/dealer.ts
Averethel: Negative. The options are typed as `{ language?: "en" | "de" | "fr" | "it" }`.
We simply can't pass more options, typechecker will forbid

Averethel: Negative, we would then end up with a `?` at the end of the url

masone: well, that's a matter of you covering for that case

Averethel: Agree, but I'd rather not couple the transformation to query string with adding the prefix that's technically not part of the query string

masone (approved):
thanks for addressing the feedback

PR #166, opened 2020-09-08
feat: rename dealerCarouselPromotion to dealerPromotionContent

ralucacril:
https://autoricardo.atlassian.net/browse/CAR-4933

removed `dealerCarouselPromotion` and using `dealerPromotionContent`

@carforyou/frontend should i release a major version or a minor should be sufficient?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,18 +41,14 @@ export {
   ListingSource,
 } from "./types/models/listing"
 
-export {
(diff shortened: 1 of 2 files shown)


Averethel (approved):
If the types are compatible `minor` if okay. Otherwise, I'd go for `major`

ralucacril:
they are compatible. thanks!

PR #170, opened 2020-09-16
feat: create buy now application call

Averethel:
An API call needed to submit the buy now form: https://autoricardo.atlassian.net/browse/CAR-4996


src/types/models/applications.ts
```diff
@@ -0,0 +1,31 @@
+export interface MoneybackApplication {
+  email?: string
+  language?: string
+  firstName?: string
+  lastName?: string
+  birthdate: string
+  street: string
+  streetNumber: string
+  city: string
+  zipCode: string
```
lkappeler, line 10: What's the reason for the zip to be a string and not a number?

lkappeler (approved):
LGTM

src/services/car/buyNowApplication.ts
```diff
@@ -0,0 +1,28 @@
...
+  buyNowApplication: BuyNowApplication,
+  { recaptchaToken = null, validateOnly = false } = {}
+): Promise<WithValidationError<BuyNowApplication>> => {
+  const path = `/listings/${listingId}/buy-now-applications${
+    validateOnly ? "/validate" : ""
+  }`
+
+  const headers = {
+    ...(recaptchaToken ? { "Recaptcha-Token": recaptchaToken } : {}),
+  }
```
masone, line 17: do we repeat this line in every API call that has a recaptcha token? would it be worth it pushing it one level down?

PR #171, opened 2020-09-17
feat: remove type from auto-alarm

ufkaya:
BREAKING CHANGE

Backend requested to remove the type: "auto-alarm"
DH update API: https://github.com/carforyou/carforyou-dealerhub-web/pull/718

### DIFF
--- a/src/lib/toQueryString.ts
+++ b/src/lib/toQueryString.ts
@@ -3,7 +3,7 @@ const toQueryString = (params: object): string => {
   return Object.keys(params)
     .filter(
       (key) =>
-        typeof params[key] !== undefined &&
+        typeof params[key] !== "undefined" &&
         typeof params[key] !== null &&
         params[key] !== ""
     )
--- a/src/services/search/autoAlarm.ts
+++ b/src/services/search/autoAlarm.ts
@@ -19,12 +19,13 @@ export const fetchSavedSearches = async (
   size?: number
 ): Promise<Paginated<DealerSavedSearch>> => {
   return withTokenRefresh(async () => {
+    const query = toQueryString({ page, size })
     try {
       const result = await fetchPath(
         Service.SEARCH,
-        `dealers/${dealerId}/listing-saved-searches/auto-alarms?${toQueryString(
-          { type: "auto-alarm", page, size }
-        )}`
+        `dealers/${dealerId}/listing-saved-searches/auto-alarms${
+          query ? `?${query}` : ""
+        }`
       )
 
       return {
--- a/src/services/search/autoAlarm.ts
+++ b/src/services/search/autoAlarm.ts
@@ -6,15 +6,15 @@ import {
   deletePath,
   postData,
 } from "../../base"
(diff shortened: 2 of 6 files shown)


src/lib/toQueryString.ts
```diff
@@ -3,7 +3,7 @@ const toQueryString = (params: object): string => {
   return Object.keys(params)
     .filter(
       (key) =>
-        typeof params[key] !== undefined &&
+        typeof params[key] !== "undefined" &&
```
masone, line 6: is this change related to removing the auto alarm type? judging from the commit history, you plan to squash-merge this, meaning it will end up in a `feat` commit

in projects, this is rarely an issue, but in packages, I'd be very explicit about the commit history. these will result in changelog entries after all. from reading the changelog, this will be an unexpected side effect

src/services/search/autoAlarm.ts
```diff
@@ -6,25 +6,26 @@ import {
...
+    const query = toQueryString({ page, size })
     try {
       const result = await fetchPath(
         Service.SEARCH,
-        `dealers/${dealerId}/listing-saved-searches/auto-alarms?type=${type}${
-          page ? `&page=${page}` : ""
-        }${size ? `&page=${size}` : ""}`
+        `dealers/${dealerId}/listing-saved-searches/auto-alarms${
+          query ? `?${query}` : ""
+        }`
```
masone, line 28: same here :)

masone (changes requested):
thanks for the valuable changes. if you clean up the commit history and rebase-merge, I'm fine with approving :) 

right now, there are two different changes in a commit history that you can't rebase, so I'm assuming you plan to squash

src/lib/toQueryString.ts
ufkaya: this is a bug that I need to fix it so I can use this function, I can add it in the feat:

src/services/search/autoAlarm.ts
ufkaya: What would be the right thing to do here? Do I need to create a PR to remove the type and another 1 to fix the bug?

src/lib/toQueryString.ts
masone: allrighty, if it's directly correlated to your code change, let's roll with squashing it

masone: but it will change the behavior for other code that consumes this function and this will not be reflected in the changelog now. this decision is ultimately up to you

src/services/search/autoAlarm.ts
masone: if they're independent of each other, that's what I would do. if you can only roll out the type when fixing the bug, I'd rewrite git history to contain two commits - one to `fix` the bug, one for the `feat` and then instead of squashing, do a rebase merge

masone (approved):
unblocking the PR now after we had the discussion on commit granularity / git history / changelog :) 
how you want to execute is up to you. there are multiple valid ways to go about it

PR #173, opened 2020-10-02
chore: update eslint config

Averethel:
We closed renovate PR so a manual update is needed.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,9 @@
 module.exports = {
   extends: ["@carforyou/eslint-config"],
   rules: {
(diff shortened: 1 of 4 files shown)


lkappeler (approved):

.eslintrc.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
   extends: ["@carforyou/eslint-config"],
   rules: {
+    // We need conditional expects for error cases in api calls
```
lkappeler, line 4: Thanks for the comment, think they are very valuable here

PR #175, opened 2020-10-05
BREAKING CHANGE: update token refresh handling

lkappeler:
References https://autoricardo.atlassian.net/browse/CAR-3971

## Motivation and context
We do encounter various issue regarding the handling of the refresh token in the dealerhub. In order to improve on the handling and avoid errors the following changes to the api client are introduced.

Reading recommendation about implementing JWT's and refresh token handling in a SSR React framework like next.js
https://hasura.io/blog/best-practices-of-using-jwt-with-graphql/

## Before

1. The api client implements a function to refresh the token. A route can be configured to request the new token; The refresh token must be set to the api client and is not updated in the consumer after issuing a new token. Besides that the route can not be fetched form the server, eg in a middleware or getServerSide props function as it needs an absolute url and a proxy configuration to make the request in the users context.

2. Parallel request produce a race condition as only the first request will be able to successfully issue a new token as the refresh token can only be used once (by best practice/config) to get a new token

3. The api client provides the `withTokenRefresh` which must be wrapped around every authenticated request to issue a new token and retry the request when the api returns a 401. This can be easily forgotten and is hard to notice while testing as your session needs to expire to get the behavior.

## After

1. The api client takes a handler function to issue a new token. This allows us in the dealerhub project to pass a handler depending on the context (client or Serverside; see [Draft PR](https://github.com/carforyou/carforyou-dealerhub-web/pull/744)) and either proxy the refresh request trough the project and use the http cookie (see linked article) or directly issue a new token form keycloak when the request fails in the servers context.   
An alternative would be to implement a proxy for all authenticated requests. This will require us to implement a handling to proxy requests on the server to this endpoint in the name of the user (eg with the user session)

2. The refresh function has been updated to store weather there is a token refresh in progress. This will prevent parallel requests form requesting multiple refresh tokens. The first request will request the new token and the successive requests will be retried after the new token has been issued.

3. I will create another pull request to update usage of the refresh function. Instead of wrapping every single request in the function I want to introduce a more imperative style where the fetch and post helper functions takes a options argument (we planned this for requesting mocked responses as well) that determines weather the call need authentication.
If a api call needs authentication two things will be added.
    1. The authentication header will be added to the request. In conclusion it won't be added if the request is not specified as authenticated, this will result in a response error. So one adding a new protected endpoint will immediately see that the authentication is missing instead of having a working request until the session expires.
    2. The token refresh handler will be wrapped around the authenticated call.

## TODO

- [x] Clean up adapt tests and docs
- [x] Add a test for the singelton instance exported
- [x] Implement parallelism for refresh handling
- [x] Go for a imperative approach on the protected endpoints ( Stacked PR )

## Update:

During the implementation in DH we discovered that initial (and current) approach has a flaw where the token might be overrriten by another users token when two requests are processed in parallel. Where the later request will override the token of the previous one. This may result in the server fetching data with the token of another user if the requests are processed in a very close timeframe. In order to avoid timing issue the api client has been refactored so the token can only be globally set when used client side. For server side requests the JWT needs to be passed for each request.

The token refresh handling will be updated to refresh the token  preemptively instead of issuing a new token on a failed request. This will eliminate the need of handling parallel requests that need a new token. Server side this should be handled by the keycloak middleware.


README.md
```diff
@@ -35,9 +35,25 @@ fetchBodyTypes()
...
 
+### Authenticated Requests
+
+#### Access Token
+
+In order to perform authenticated requests you need to set a valid access token to the api client in the consumers code `ApiClient.setAccessToken(JWT)` The provided token will be passed as `Authentication` header to the API.
+
+#### Refreshing Access Tokens
+
+Authenticated requests will try to issue a refresh token when they revive a 401 response form the remote server. In order to handle the process of refreshing the token the consumer must pass a handler function to the API-Client. (The function may vary depending on the environment the API-Client is used)
```
masone, line 48:
```suggestion
Authenticated requests will try to issue a refresh token when they receive a 401 response form the remote server. In order to handle the process of refreshing the token the consumer must pass a handler function to the API-Client. (The function may vary depending on the environment the API-Client is used)
```

README.md
```diff
@@ -35,9 +35,25 @@ fetchBodyTypes()
...
+
+```
+ApiClient.setTokenRefreshHandler(async () => {
+  const response = await fetch(process.env.TOKEN_REFRESH_ROUTE)
+  return response.json()
+})
+```
```
masone, line 55: I might see the answer later in the code, but how is the case handled when this function is missing, but the API client runs into a 401. A helpful message would be appreciated over throwing a cryptic message :)

masone: will we have spot-on types for this functions return value to make sure we don't fuck refreshing up?

src/__tests__/tokenRefresh.test.ts
```diff
@@ -2,108 +2,151 @@ import apiClient from "../apiClient"
...
+      expect(mockTokenRefreshHandler).toHaveBeenCalledTimes(0)
+    })
+
+    it("rethrows the error", async () => {
+      await expect(withTokenRefresh(apiCall)).rejects.toThrowError(
+        new ResponseError({ status: 400 })
+      )
     })
   })
 })
```
masone, line 152: nice job on the tests here! 💯

src/apiClient.ts
```diff
@@ -6,30 +6,37 @@ export interface ApiClientConfig {
...
-  handlers: Handlers = {}
-  version = "v1"
+  configuration: ApiClientConfig
+  accessToken: string | null
+  handlers: Handlers
+  version: string
+  refreshToken: () => Promise<{ accessToken: string }>
 
   constructor() {
+    if (ApiClient.instance) return ApiClient.instance
```
masone, line 27: could I have instantiated two api clients with different configurations before and can I still do this now? I'm not asking for that capability, but it might be worth thinking about whether we'd ever want this. or at least it's a breaking change worth pointing out

src/apiClient.ts
```diff
@@ -6,30 +6,37 @@ export interface ApiClientConfig {
...
+
+    this.configuration = {}
+    this.accessToken = null
+    this.handlers = {}
+    this.version = "v1"
+    this.refreshToken = async () => {
+      throw new Error(
+        "The refreshToken function has not been set the the apiClient instance, use apiClient.setTokenRefresh to pass a function which handles the token refresh"
```
masone, line 35:
```suggestion
        "The refreshToken function has not been set the the apiClient instance, use apiClient.setTokenRefreshHandler to pass a function which handles the token refresh"
```

src/apiClient.ts
```diff
@@ -42,9 +49,22 @@ class ApiClient {
...
+      configured: Object.keys(this.configuration).length > 0,
+      version: this.version,
+    }
+  }
+
+  public setAccessToken(accessToken: string): void {
+    this.accessToken = accessToken
+  }
+
+  public setTokenRefreshHandler(
```
masone, line 64: what's the motivation to have `setTokenRefreshHandler` as a function on the singleton over passing it with the configuration initially?

src/index.ts
```diff
@@ -1,4 +1,6 @@
-import apiClient, { Handlers, Tokens, ApiClientConfig } from "./apiClient"
+import apiClient from "./apiClient"
+// TODO: do we really wanna handle it like this and export the instance as if is a constructor
```
masone, line 2: TODO

masone: I'm ok with this :)

masone: maybe this refers to my comment above? https://github.com/carforyou/carforyou-api-client-pkg/pull/175/files#diff-1c1a34a94ec474ad2092388c3650708eR27

src/tokenRefresh.ts
```diff
@@ -1,41 +1,41 @@
...
-    refreshToken: refresh_token,
-  })
+
+const invokeTokenRefresh = async () => {
+  const { accessToken } = await apiClient.refreshToken()
+
+  apiClient.setAccessToken(accessToken)
   if (apiClient.handlers.onAccessTokenUpdate) {
-    apiClient.handlers.onAccessTokenUpdate(apiClient.tokens.accessToken)
+    apiClient.handlers.onAccessTokenUpdate(apiClient.accessToken)
```
masone, line 8: is this something the following code needs to wait for to prevent race conditions? like... the token is updated in the client, but not yet via the callback in the project

src/tokenRefresh.ts
```diff
@@ -1,41 +1,41 @@
...
 }
 
+let pendingRefreshRequest = false
+const waitForPendingRefresh = async (apiCall) => {
+  return new Promise((resolve) => {
+    setTimeout(async () => {
+      return pendingRefreshRequest
+        ? resolve(waitForPendingRefresh(apiCall))
+        : resolve(apiCall())
+    }, 100)
```
masone, line 19: can there be a case where this causes an endless loop or a very long running one that we'd want to time out at some point?

src/tokenRefresh.ts
```diff
@@ -1,41 +1,41 @@
...
+  })
+}
+
 const refreshTokenAndRetry = async (apiCall) => {
+  if (pendingRefreshRequest) return waitForPendingRefresh(apiCall)
+
   try {
-    await refreshToken()
-    return apiCall()
+    pendingRefreshRequest = true
```
masone, line 27:
```suggestion
    hasPendingRefreshRequest = true
```

to indicate with the name that it's a "state" and not a "request"?

masone: really optional. the types might point this out already

src/tokenRefresh.ts
```diff
@@ -1,41 +1,41 @@
...
-      if (
-        (status.startsWith("4") || status.startsWith("5")) &&
-        apiClient.handlers.onFailedTokenRefresh
-      ) {
-        apiClient.handlers.onFailedTokenRefresh()
-      }
-
-      return
+    if (apiClient.handlers.onFailedTokenRefresh) {
+      apiClient.handlers.onFailedTokenRefresh()
```
masone, line 32: similar to above - do we need to await this?

masone (commented):
first round of clarifying questions from my side. looks like you're 99% there, but don't feel comfortable approving just yet without understanding some details :)

README.md
lkappeler: 1. My intention was to throw a meaningful error in the function added in the prototype https://github.com/carforyou/carforyou-api-client-pkg/pull/175/files#diff-c50cd8cbf813f34536ae74335bcf5724R56 so if one misses to add it the error shall point them to whats off

2. The function is typed pretty strict https://github.com/carforyou/carforyou-api-client-pkg/pull/175/files#diff-1c1a34a94ec474ad2092388c3650708eR65 let me know if you would add something on that.

src/apiClient.ts
lkappeler: I am not sure whether I understood the intention on the existing implementation properly but I assumed that having a singelton for the api client is intended and there was a flaw in the implementation
Maybe @Averethel can clarify on this?

src/apiClient.ts
lkappeler: My though behind was that the configuration is something that can be serialized to a static file eg. loading it form the .env files where as the handler needs to be passed as function that may need application context...
If you check the implementation in DH https://github.com/carforyou/carforyou-dealerhub-web/pull/744/files#diff-bc6a44fc684b34709ef25152c25c9b15R9 the configuration can just be "statically" loaded on the module top level where as the refresh handler needs to be set in the middleware to have the proper context

src/tokenRefresh.ts
lkappeler: Not sure if I understand you correctly this is the client callback executor

src/apiClient.ts
Averethel: We still can have a singleton in a node process with a server-side configuration (think pre-fetching) and a client-side one, right?

src/tokenRefresh.ts
lkappeler: Ahh gotcha in the other comment will check it

src/tokenRefresh.ts
lkappeler: Fair point, will update it manually to keep the code running

src/tokenRefresh.ts
```diff
@@ -1,41 +1,41 @@
...
+
+  apiClient.setAccessToken(accessToken)
   if (apiClient.handlers.onAccessTokenUpdate) {
-    apiClient.handlers.onAccessTokenUpdate(apiClient.tokens.accessToken)
+    apiClient.handlers.onAccessTokenUpdate(apiClient.accessToken)
   }
 }
 
+let pendingRefreshRequest = false
+const waitForPendingRefresh = async (apiCall) => {
```
Averethel, line 13: Would it make sense to limit the retries here?

Averethel (approved):
Looking solid overall, I added a bunch of comments on top of what @masone asked.
Good job

README.md
masone: thanks for double checking. I forgot to delete this reminder-to-self-comment when I reached the function and answered the question to myself ;)

src/apiClient.ts
masone: yes, that's what I see too

src/tokenRefresh.ts
masone: that said - you're in dangerous territory here. these things are hard to test and hard to detect in production. if you have a simpler approach, I'd prefer that over potential endless loop

src/base.ts
```diff
@@ -79,38 +75,73 @@ export const resolveServiceUrl = (service: Service): string => {
...
 }
 
-export const fetchPath = async (
-  service: Service,
-  path: string,
-  options = {}
-) => {
-  const prefix = resolveServiceUrl(service)
-  const url = `${prefix}/${stripLeadingSlash(path)}`
+export const fetchPath = async ({
```
lkappeler, line 97: Went with updating the interface over a polymorphic options strategy where the options could be a different arguement based on weather the current request has a body or not

README.md
```diff
@@ -35,9 +35,33 @@ fetchBodyTypes()
 | `optionServiceUrl`           | URL to [Options Service](https://option-service.preprod.carforyou.ch/swagger-ui/index.html)                         |
 | `analyticsServiceUrl`        | URL to [Analytics Service](https://carforyou-analytics-service.preprod.carforyou.ch/swagger-ui/index.html)          |
 | `userNotificationServiceUrl` | URL to [User Notification Service](https://user-notification-service.preprod.carforyou.ch/swagger-ui/index.html)    |
-| `tokenRefreshServiceUrl`     | URL to Auth Service used to refresh access tokens                                                                   |
 | `debug`                      | Set to `true` to `console.log` requests and API responses.                                                          |
 
+### Authenticated Requests
+
+Authenticated requests need to be handled in a different way depending on weather the request is performed server- or client-side.
```
masone, line 42:
```suggestion
Authenticated requests need to be handled in a different way depending on whether the request is performed server- or client-side.
```

src/apiClient.ts
```diff
@@ -6,30 +6,55 @@ export interface ApiClientConfig {
...
+    this.runsInBrowser = typeof window !== "undefined"
+    this.refreshToken = async () => {
+      throw new Error(
+        "The refreshToken function has not been set the the apiClient instance, use apiClient.setTokenRefreshHandler to pass a function which handles the token refresh"
+      )
+    }
+    this.ensureTokenFreshness = async () => {
+      console.log("ensure fresshenes!!!", Date.now())
+      console.log("ensure fresshenes!!!", this.accessToken)
+      // TODO: check weather the token is n minutes valid if not refreh it
```
masone, line 50: todo and logs before merging
```suggestion
      // TODO: check whether the token is n minutes valid if not refresh it
```

README.md
```diff
@@ -35,9 +35,33 @@ fetchBodyTypes()
...
+Client-Side, we can use the api client as a singelton to store the token. Server-side the access token needs to be passed to the request in order
```
masone, line 43:
```suggestion
Client-Side, we can use the api client as a singleton to store the token. Server-side the access token needs to be passed to the request in order
```

I was involved in this decision, but reading this, I'm thinking - but why! Is it worth putting the mental overhead on the consumer to decide which one is the right approach for what context?
How much worse would it be for a server-side consumer if they'd have to pass the token with every call as well? The app can still decide to store the token in some singleton-ish way

README.md
```diff
@@ -35,9 +35,33 @@ fetchBodyTypes()
...
+to not expose user credentials to the global module scope and ensure the request is performed with the correct
```
masone, line 44: incomplete sentence?

README.md
```diff
@@ -35,9 +35,33 @@ fetchBodyTypes()
...
+
+#### Implementing authenticated requests
+
+To implement a authenticated request, simply pass the `isAuthorizedRequest` option to the request helper (fetch/post/put/delete-Data) and the api client will add the `Authorization` header. For server side requests, don't forget to pass the options down to the helper function so the consumer can pass the `accessToken` when invoking the request.
```
masone, line 48:
```suggestion
To authenticate a request, simply pass the `isAuthorizedRequest` option to the request helper (fetch/post/put/delete-Data) and the api client will add the `Authorization` header. For server side requests, don't forget to pass the options down to the helper function so the consumer can pass the `accessToken` when invoking the request.
```

masone: reading this, I'd love to get rid of all the "if this then that"'s (refers to the comment above). maybe we can simplify server/client side handling?

I'm also a bit confused by passing `isAuthorizedRequest`. wouldn't we just pass the token and that would then authenticate the request?

README.md
```diff
@@ -35,9 +35,33 @@ fetchBodyTypes()
...
+
+#### Access Token Handling
+
+##### Client-Side requests
+
+Set the access token along with it's expiration date to the api client using the `setToken` function provided by the api client instance.
+
+##### Server-Side requests
+
+Pass the access token as a request option to the api call `dummyFetchCall(data, {accessToken: JWT})`
```
masone, line 58: wouldn't that be the client-side version? the title of this paragraph says it's for server-side

src/__tests__/base.test.ts
```diff
@@ -39,13 +39,14 @@ describe("Base", () => {
     })
   })
 
+  // TODO: add some tests for thems new interface
```
masone, line 42: todo

src/__tests__/tokenRefresh.test.ts
```diff
@@ -1,109 +1,152 @@
...
-
-      it("rethrows the error", async () => {
-        await expect(
-          withTokenRefresh(apiCall)
-        ).rejects.toThrowErrorMatchingSnapshot()
-      })
-    })
-  })
-})
+// import apiClient from "../apiClient"
```
masone, line 1: todo: delete

src/apiClient.ts
```diff
@@ -6,30 +6,55 @@ export interface ApiClientConfig {
...
-  configuration: ApiClientConfig = {}
-  tokens: Tokens = {}
-  handlers: Handlers = {}
-  version = "v1"
+  configuration: ApiClientConfig
+  accessToken: {
+    token: string | null
+    expires: Date | null
+  }
+  handlers: Handlers
```
masone, line 25: I could imagine not grouping these in a `handlers` variable. passing `onFailedTokenRefresh` directly would make the interface clearer to me. I had a "what are handlers" moment and had to look at the types to figure out what to do

really optional - feel free to go with this if you have good reasons to

src/apiClient.ts
```diff
@@ -6,30 +6,55 @@ export interface ApiClientConfig {
...
+    if (ApiClient.instance) return ApiClient.instance
+
+    this.configuration = {}
+    this.accessToken = {
+      expires: null,
+      token: null,
+    }
+    this.handlers = {}
+    this.version = "v1"
+    this.runsInBrowser = typeof window !== "undefined"
```
masone, line 41:
```suggestion
    this.hasWindowObject = typeof window !== "undefined"
```

if you want to harmonize with the naming in the deviceHelper in listings

src/apiClient.ts
```diff
@@ -6,30 +6,55 @@ export interface ApiClientConfig {
...
+    this.refreshToken = async () => {
+      throw new Error(
+        "The refreshToken function has not been set the the apiClient instance, use apiClient.setTokenRefreshHandler to pass a function which handles the token refresh"
```
masone, line 44:
```suggestion
        "The refreshToken function has not been set on the apiClient instance, use apiClient.setTokenRefreshHandler to pass a function which handles the token refresh"
```

src/apiClient.ts
```diff
@@ -42,9 +67,33 @@ class ApiClient {
...
   }
 
-  public setTokens(tokens: Tokens): void {
-    this.tokens.accessToken = tokens.accessToken
-    this.tokens.refreshToken = tokens.refreshToken
+  public getConfiguration() {
+    return {
+      ...this.configuration,
+      configured: Object.keys(this.configuration).length > 0,
+      version: this.version,
```
masone, line 74: can this default be merged when the api client is initialized? looking at `this.configuration` at runtime might yield a different result as when calling `getConfiguration`. we're opening ourselves to potential mishaps with this

src/apiClient.ts
```diff
@@ -42,9 +67,33 @@ class ApiClient {
     })
   }
 
-  public setTokens(tokens: Tokens): void {
-    this.tokens.accessToken = tokens.accessToken
-    this.tokens.refreshToken = tokens.refreshToken
+  public getConfiguration() {
+    return {
+      ...this.configuration,
+      configured: Object.keys(this.configuration).length > 0,
```
masone, line 73: what is this used for? seems like an odd thing to add to the configuration itself. if we need to perform this check somewhere (can't find where in the github diff), I'd move it there instead

src/apiClient.ts
```diff
@@ -42,9 +67,33 @@ class ApiClient {
...
+      version: this.version,
+    }
+  }
+
+  public setAccessToken(accessToken: { token: string; expires: Date }): void {
+    if (!this.runsInBrowser) {
+      throw new Error(
+        "You may only globally set and access-token on a client side usage, to avoid leaking the token on the global module for a server side implementation. Please pass tha access-token as an option to the request instead when using the package server side!"
```
masone, line 81:
```suggestion
        "You may only globally set (xyz) and access-token on a client side usage, to avoid leaking the token on the global module for a server side implementation. Please pass the access-token as an option to the request instead when using the package server side!"
```

src/base.ts
```diff
@@ -132,36 +163,60 @@ export const fetchPath = async (
...
-    method: "PUT",
+export const putData = async ({
+  service,
+  path,
+  body,
+  options = {},
+}: {
+  service: Service
+  path: string
+  body: unknown
```
masone, line 196: what's the consequence of changing `object` to `unknown`?

src/tokenRefresh.ts
```diff
@@ -1,51 +1,52 @@
...
-  } catch (error) {
-    if (error.name === "ResponseError" && error.status.toString() === "401") {
-      return refreshTokenAndRetry(apiCall)
-    }
-
-    throw error
-  }
-}
+
+// const invokeTokenRefresh = async () => {
```
masone, line 3: todo delete file

masone (changes requested):
Thanks for putting the time understanding the details of our current implementation. I think the overall approach is absolutely sane. Now that I see it in action, I question our call for handling client/server side differently when it comes to passing the token. I feel always passing the token would be much clearer. It might be a question of how hard this is to do in the app though - up for another tradeoff discussion? you might have the answer to that handy already

src/apiClient.ts
lkappeler: Not sure where the function is consumed but kept it as is, just moved it form the index to the class as I removed the wrapper....
Krzysztof will tackle the configuration in his PR so it might be a good opportunity to update there...

src/apiClient.ts
lkappeler: DH consumes it to check weather to re-configure the api client or not.....

src/__tests__/base.test.ts
```diff
@@ -55,26 +55,101 @@ describe("Base", () => {
...
+      it("does not throw and error when the request is not marked as authorized", async () => {
+        const response = await fetchPath({
+          service: Service.CAR,
+          path: "api/path",
+          options: {
+            accessToken: "GIMME ACCESS!",
+          },
+        })
+        expect(response).toEqual({ ok: true })
+      })
```
Averethel, line 101: Shouldn't this test not include `accessToken` in the options?

Averethel: As in the way I understand it you `throw` when the request is marked as `isAuthorized` and there's no token, right?

src/__tests__/base.test.ts
```diff
@@ -55,26 +55,101 @@ describe("Base", () => {
...
+          expect.any(String),
+          expect.objectContaining({
+            headers: expect.objectContaining({
+              Authorization: "Bearer GIMME ACCESS!",
+            }),
+          })
+        )
+      })
+
+      it("throws an error if the toke is not passed for an authenticated request", async () => {
```
Averethel, line 141:
```suggestion
      it("throws an error if the token is not passed for an authenticated request", async () => {
```

src/base.ts
```diff
@@ -79,38 +75,68 @@ export const resolveServiceUrl = (service: Service): string => {
...
-  return `Bearer ${apiClient.tokens.accessToken}`
+  return { Authorization: `Bearer ${accessToken}` }
+}
+
+export interface RequestOptions {
+  isAuthorizedRequest?: boolean
+  accessToken?: string
+  method?: "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
+  headers?: HeadersInit
 }
```
Averethel, line 93: In my PR I created two interfaces, one private for fetch wrappers and one public for the calls: https://github.com/carforyou/carforyou-api-client-pkg/pull/179/files#diff-440b55a9ff9e0486020fc49197f3278d14446306fc0c5d5107b48139c20bfda6R41-R51

This might be a good idea here too. Otherwise you open yourself to a situation where a consumer can call:

`getListing({ id: 123, method: "POST" })` or `getDealerListings({ dealerId: 123, isAuthorizedRequest: false })`

Averethel (commented):
Looks good to me. Take a look at the comment on the options interfaces since it creates some potential loopholes

src/__tests__/base.test.ts
lkappeler: I can remove it if it makes it more clear, left it there on purpose as it should not matter weather you pass it or not for a public call...

src/base.ts
lkappeler: I thought about that too and went with a unified interface to make it simpler to work with as you don't need to determine between internal and external calls and it provides the max amount of flexibility. Yes someone could fuck up the request like that but I expect some logical sense when consuming it to not switch pre configured settings. There might be cases where it gets handy to be have to override them for test cases or so if you are working with mock responses and don't need to pass tokens so you could tweak the behavior to not throw for those cases...

I don't mind switching in you PR, do you think we need to do the same "twice" or will it be fine if we just split the interfaces in your PR?

README.md
```diff
@@ -35,9 +35,26 @@ fetchBodyTypes()
...
+
+Note: If one forgets to add the `isAuthorizedRequest` option, the access token will not be set as a header. On the consumer side, if the access token is not passed, the api client will throw an error.
+
+#### Performing authenticated requests
+
+Pass the access token as a request option to the api call `dummyFetchCall(data, {accessToken: JWT})`
+
+#### Refreshing Access Tokens
+
+The consumer is responsible to ensure a valid token is passed to the request. The api client will pass the provided token as an `Authorization` header to the api call.
```
masone, line 56: if you have the energy, it might be worth documenting how to do that. can be separate, no need to rush it

src/apiClient.ts
masone: cc @Averethel

src/base.ts
```diff
@@ -79,38 +75,68 @@ export const resolveServiceUrl = (service: Service): string => {
   throw new Error(`Missing endpoint configuration for "${service}" service`)
 }
 
-const authorizationHeader = () => {
-  if (!apiClient.tokens.accessToken) {
-    return null
+const getAuthorizationHeader = (accessToken = null) => {
```
masone, line 78:
```suggestion
const getAuthorizationHeader = (accessToken) => {
```

results in the same behavior :)

src/base.ts
```diff
@@ -79,38 +75,68 @@ export const resolveServiceUrl = (service: Service): string => {
...
-  const prefix = resolveServiceUrl(service)
-  const url = `${prefix}/${stripLeadingSlash(path)}`
+export const fetchPath = async ({
+  service,
+  path,
+  body,
+  options = {},
+}: {
+  service: Service
+  path: string
```
masone, line 102: I would keep service/path as regular attributes - but really up to you. I think krzysztof is preferring named attributes as well

src/services/car/inventoryData.ts
```diff
@@ -1,49 +1,55 @@
...
 }
 
 export const fetchOptions = (): Promise<string[]> => {
-  return fetchPath(Service.CAR, "inventory/options")
+  return fetchPath({ service: Service.CAR, path: "inventory/options" })
 }
 
 export const fetchSeats = (): Promise<number[]> => {
-  return fetchPath(Service.CAR, "inventory/seats")
+  return fetchPath({ service: Service.CAR, path: "inventory/seats" })
```
masone, line 47: this diff illustrates pretty well how much more code we overall generate to achieve the same as with regular attributes :)

masone (approved):
I have reviewed this before and felt comfortable only giving it a quick peek. It looks overall sane. No blockers from my side. Really nice work!

src/base.ts
lkappeler: The reason I went with this is there is an optional body attribute in the current interface on position 3 adding the options will result in a polymorpic interface where the options might be argument 3 or 4 have seen this in framework code where they try to verify that the last argument contains the options why I don't like a lot....
Further on it can be handy if you change the interface eg Krzysztof will probably drop the service so we don't need to shift argument; Same applies for extending it

src/base.ts
lkappeler: I don't like undefined 😉 😂 
It's somewhat a mix between personal preference and readability whereas for me this indicates that the function might be called without the parameter and it does some handling in that case... Compared to not defaulting it will result in reading the function but that might be just me

PR #178, opened 2020-10-13
feat: add listings and leads analytics api

ufkaya:
[CAR-5242 - Cockpit: listings and lead number](https://autoricardo.atlassian.net/browse/CAR-5242)

### DIFF
--- a/src/services/analytics.ts
+++ b/src/services/analytics.ts
@@ -1,9 +1,6 @@
 import { postData, Service } from "../base"
 
-import {
-  DealerListingsAnalyticsData,
-  ListingsLeadsAnalytics,
-} from "../types/models"
+import { DealerListingsAnalyticsData, CockpitAnalytics } from "../types/models"
 
 export const fetchAnalyticsData = async (
   dealerId: number,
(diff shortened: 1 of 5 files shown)


src/services/analytics.ts
```diff
@@ -14,3 +18,32 @@ export const fetchAnalyticsData = async (
...
+    query,
+  })
+}
+
+export const fetchListingsAnalytics = async ({
+  dealerId,
+  dimensions,
+}: {
+  dealerId: number
+  dimensions?: string[]
```
masone, line 43: is it worth typing the dimensions here? or is it an undefined set of strings?

src/services/analytics.ts
```diff
@@ -14,3 +18,32 @@ export const fetchAnalyticsData = async (
...
 }
+
+export const fetchLeadsAnalytics = async ({
+  dealerId,
+  dimensions,
+  query,
+}: {
+  dealerId: number
+  dimensions?: string[]
+  query?: { verificationDateFrom: string }
```
masone, line 29: is this the only query parameter available on this call?

masone (approved):

src/services/analytics.ts
ufkaya: I need it in case we want to pass the dimension like `conditionType`

masone:
@Averethel @lkappeler you'll need to rebase on these :)

lkappeler:
Jup did that

PR #179, opened 2020-10-14
feat: overwrite services per call

Averethel:
**BREAKING CHANGE**

This PR introduces the following:
- ensures that all the API call take named parameters instead of positional parameters
- ensures that every call accepts options objects which can be used to:
  - set the headers per call
  - overwrite the service URL for the specific call
     as a consequence of this configuration is simplified and only API gateway URL and refresh token URL are necessary
- renames a couple of API calls that had typos in names

While this is still a draft and some work is required early feedback would be very welcome

TODO:
- [x] unify/align named parameters across API calls
- [x] add tests for previously untested calls
- [x] align this with refresh token handling changes
- [x] update documentation


src/base.ts
```diff
@@ -41,76 +30,69 @@ const withPagination = <T>(json: {
...
-    return url
-  }
+  return `Bearer ${apiClient.tokens.accessToken}`
+}
 
-  throw new Error(`Missing endpoint configuration for "${service}" service`)
+export interface RequestOptions {
+  headers?: Record<string, string>
+  // TODO: think about this naming
+  serviceUrl?: string
```
masone, line 44:
```suggestion
  host?: string
```

src/base.ts
```diff
@@ -41,76 +30,69 @@ const withPagination = <T>(json: {
...
+}
+
+type Method = "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
+
+interface RequestOptionsWithMethod extends RequestOptions {
+  method?: Method
+}
+
+export const getHost = ({ serviceUrl }) => {
+  const url = serviceUrl || apiClient.configuration.apiGatewayUrl
```
masone, line 58: I noticed inconsistent naming - method is called `getHost`, but the variables `serviceUrl` and `url`

src/base.ts
```diff
@@ -41,76 +30,69 @@ const withPagination = <T>(json: {
...
-  const prefix = resolveServiceUrl(service)
-  const url = `${prefix}/${stripLeadingSlash(path)}`
+export const fetchPath = async ({
+  path,
+  body,
+  options,
+}: {
+  path: string
+  body?: string
+  options: RequestOptionsWithMethod
```
masone, line 76: 1. I really liked to have path as the first (unnamed) argument as it's the only required one. the majority of simple GET would be a tad more straight forward

2. would options be optional?

src/base.ts
```diff
@@ -41,76 +30,69 @@ const withPagination = <T>(json: {
...
   const response = await fetch(url, {
     headers: {
       "Content-Type": "application/json",
       Accept: `application/vnd.carforyou.${apiClient.version}+json`,
       ...(auth ? { Authorization: auth } : {}),
       ...headers,
     },
-    ...otherOptions,
+    method,
+    ...(body ? { body } : {}),
```
masone, line 95: this might be intentional, but worth pointing out - this removes the possibility to pass arbitrary options to fetch. we're locking down to headers, method and body. if it was intentional, please note it in the breaking changes and verify we're not making use of that anywhere

src/base.ts
```diff
@@ -132,36 +114,79 @@ export const fetchPath = async (
...
+  const { recaptchaToken, headers, ...otherOptions } = options
+
+  return {
+    ...otherOptions,
+    headers: {
+      ...headers,
+      ...(recaptchaToken ? { "Recaptcha-Token": recaptchaToken } : {}),
+    },
+    method,
+  }
```
masone, line 133: there is some header merging going on in `fetchPath` as well. worth to harmonize or a completely different thing?

src/index.ts
```diff
@@ -231,7 +235,7 @@ export { Options as OptionsFactory } from "./lib/factories/options"
 export { withRetries } from "./lib/withRetries"
 
 export const ApiClient = {
-  configure: (configuration: ApiClientConfig) =>
+  configure: (configuration: ApiClientConfiguration) =>
```
masone, line 238: in case we expose this type, please note the breaking change in the release notes

src/services/__tests__/catalogue.test.ts
```diff
@@ -36,7 +36,7 @@ describe("CATALOGUE service", () => {
     })
 
     it("returns data", async () => {
-      const fetched = await fetchModels("audi")
+      const fetched = await fetchModels({ makeKey: "audi" })
```
masone, line 39: I liked the unnamed version of this. not sure if it's personal preference, but when there is only 1 argument and it's a required one, I'd not go for named attributes

src/services/__tests__/dealer.test.ts
```diff
@@ -13,32 +13,32 @@ describe("Dealer", () => {
   describe("#fetchDealerSuggestions", () => {
     it("encodes the query", async () => {
       const query = "k+k"
-      await fetchDealerSuggestions(query)
+      await fetchDealerSuggestions({ query })
```
masone, line 16: yeah, with pretty much all of these, I struggle seeing the value in the change. can we get some other opinions?

src/services/__tests__/catalogue.test.ts
masone: also, this breaking change is pretty hefty. we'll need to update pretty much all existing calls. I don't see the value in exercising through that

src/services/car/inventory.ts
```diff
@@ -31,29 +31,44 @@ const sanitizeListing = (json): Listing => {
...
 
-export const fetchDealerMakes = async (
+export const fetchDealerMakes = async ({
+  dealerId,
+  options = {},
+}: {
   dealerId: number
-): Promise<Array<{ make: string; makeKey: string }>> => {
-  return fetchPath(Service.CAR, `inventory/dealers/${dealerId}/makes`)
+  options?: RequestOptions
```
masone, line 51:
```suggestion
  options?: ApiCallOptions
```

I just realized we never use the word `request` in this context so far. maybe align with existing naming?

src/base.ts
```diff
@@ -41,76 +30,69 @@ const withPagination = <T>(json: {
...
+  recaptchaToken?: string
+}
+
+type Method = "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
+
+interface RequestOptionsWithMethod extends RequestOptions {
+  method?: Method
+}
+
+export const getHost = ({ serviceUrl }) => {
```
masone, line 57:
```suggestion
export const getHost = (serviceUrl) => {
```

keep it simple. no need to reach for an options object when there is only one attribute

__tests__/setup/setupConfiguration.ts
```diff
@@ -2,13 +2,7 @@ import ApiClient from "../../src/apiClient"
...
 beforeAll(() => {
   ApiClient.configure({
-    carServiceUrl: "car.service.test",
-    catalogueServiceUrl: "catalogue.service.test",
-    dealerServiceUrl: "dealer.service.test",
-    searchServiceUrl: "search.service.test",
-    optionServiceUrl: "option.service.test",
-    analyticsServiceUrl: "analytics.service.test",
-    userNotificationServiceUrl: "user_notification.service.test",
+    apiGatewayUrl: "test.gateway",
```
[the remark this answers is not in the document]
Averethel: Good suggestion! I was struggling to find the right name and this is so simple.

src/base.ts
Averethel: 1. This is in line with what @lkappeler is doing for token handling change (https://github.com/carforyou/carforyou-api-client-pkg/pull/175/files#diff-440b55a9ff9e0486020fc49197f3278d14446306fc0c5d5107b48139c20bfda6R95). I'd like to keep it consistent to minimize the conflicts.

2. I thought about it and opted not to make it optional for the internal methods. My reasoning is that this way we ensure (via type safety) that we include and pass options for the new API calls down to wrappers around fetch since new logic of overwriting gateway URL relies on that.

src/base.ts
Averethel: I'll take a look at the usage in the projects. I don't think we used that but if it's something we desire and actively use it should be easy to get back.

src/base.ts
Averethel: Oh, good point. We can move it to the top-level fetch call and we don't need to worry about recaptcha token anywhere.

src/index.ts
Averethel: We don't. This is probably also changing with the token refresh I'd double check when @lkappeler merges.

src/services/__tests__/catalogue.test.ts
Averethel: I prefer to have them aligned and don't have to think if I need to pass an object or not. This also makes extending API calls easy. Plus you know immediately what parameter you pass does. It's fairly similar to always forcing parens by prettier in that regard in my opinion.
I'll take care of the update. typecheck task will do a good job guiding me through it so I'm not concerned here

src/services/__tests__/dealer.test.ts
Averethel: Let's keep the discussion about it here: https://github.com/carforyou/carforyou-api-client-pkg/pull/179#discussion_r505456515

src/base.ts
Averethel: Was low effort so I did it anyway

Averethel:
@masone ready for round 2

src/base.ts
```diff
@@ -41,76 +30,79 @@ const withPagination = <T>(json: {
...
+}: {
+  path: string
+  body?: string
+  options: RequestOptions
+}) => {
+  const {
+    method = "GET",
+    host = null,
+    headers,
+    recaptchaToken,
```
lkappeler, line 91: We could just create a helper where we use it and pass it as options.headers (Up to you would assume this to be a rather rare case just for form requests)

src/base.ts
```diff
@@ -41,76 +30,79 @@ const withPagination = <T>(json: {
...
+    ...fetchOptions
```
lkappeler, line 92: I checked this in my PR and we do not use it in the code as far as I can tell so why not just remove it and be explicit if we need a specific feature in the future and add the used feature instead of "whitelisting" a lot of the default interface

src/__tests__/base.test.ts
```diff
@@ -55,36 +42,75 @@ describe("Base", () => {
...
+            "Recaptcha-Token": "token",
+          },
+        })
+      )
+    })
+
+    it("passes fetch options to fetch", async () => {
+      const json = await fetchPath({
+        path: "/api/path",
+        options: { mode: "no-cors" },
```
lkappeler, line 95: Do we really need that we only make use of it in this test case but not in the code itself...

src/base.ts
```diff
@@ -41,76 +30,79 @@ const withPagination = <T>(json: {
...
   }
 
-  if ((url || "").trim()) {
-    return url
-  }
+  return `Bearer ${apiClient.tokens.accessToken}`
+}
 
-  throw new Error(`Missing endpoint configuration for "${service}" service`)
+export interface ApiCallOptions extends Omit<RequestInit, "method" | "body"> {
```
lkappeler, line 41: What's the reasoning behind naming Request options Call?

src/base.ts
```diff
@@ -132,36 +124,60 @@ export const fetchPath = async (
...
+}: {
+  path: string
   // eslint-disable-next-line @typescript-eslint/ban-types
-  body: object,
-  headers = {}
-) => {
-  return fetchPath(service, path, {
-    method: "POST",
+  body: object
+  options: ApiCallOptions
```
lkappeler, line 135: I might miss something but isn't the base file providing internal interfaces so it should be consistent for all helpers provided by this file?

src/services/__tests__/dealer.test.ts
```diff
@@ -13,32 +13,32 @@ describe("Dealer", () => {
...
       expect(fetch).toHaveBeenCalledWith(
-        `dealer.service.test/dealers/123?language=de`,
+        expect.stringContaining("dealers/123?language=de"),
         expect.any(Object)
       )
     })
 
     it("works without the language option", async () => {
-      await fetchDealer(123)
+      await fetchDealer({ id: 123 })
```
lkappeler, line 38: it might be a bit arbitrary so up to you but since you are on it it might be helpful to name the id params like `dealerId` and `listingsId` (we might have other ids) so it would be more obvious in the consumer code what ids are passed around

lkappeler: Just saw below that you already named the other ones so it probably make sense to be consistent

src/services/car/moneybackApplication.ts
```diff
@@ -1,30 +1,43 @@
...
-    validateFields: Object.keys(moneybackApplication),
-    ...options,
+export const sendMoneybackApplication = async ({
+  listingId,
+  moneybackApplication,
+  options = {},
+}: {
+  listingId: number
+  moneybackApplication: MoneybackApplication
+  options: ApiCallOptions & {
```
lkappeler, line 13: I would avoid combining request options with sepecific options for the concrete endpint

src/types/params/index.ts
```diff
@@ -2,3 +2,5 @@ export interface PaginationParams {
   page?: number
   size?: number
 }
+
+export type Language = "de" | "en" | "fr" | "it"
```
lkappeler, line 6: I would not export this form the api client and use it as an internal interface as the package should not defined languages or name it specifically for the use case as we use Lanuges for the applications language support....

lkappeler (commented):
Looking good so far, comments up for discussion. I think it's fine to unify the interfaces as we probably will benefit form it in the future when we extend or change existing requests. Up to you whether you wanna bundle that with this change or split it out

On the general interface I would split the paramters for the concrete request form generic request options ( commented that somewhere) which will result in something like this 
```

interface ConcreteRequestParameters {
    dealerId: number,
    listingId: number,
    options: {
        validateOnly: boolean
    }
}

interface RequestOptions {
    headers: Headers
    accessToken: string
    recaptchaToken: string
    method: "GET" | "POST"
}

const fetchRequest({ dealerId, listingId, options: {validateOnly}} : ConcreteRequestParameters, options: RequestOptions) {
    //...
}
```

It might be worth considering something like to imply the handling as in custom ones will be manipulated and the rest will be passed to the fetch function (did not think of that in my PR...)
```
interface RequestOptions {
    headers: Headers
    method: "GET" | "POST"
    custom: {
        recaptchaToken: string
        accessToken: string
    }
}
```

src/base.ts
Averethel: I like passing it directly and having the client taking care of the header logic. This way the header name is local to the client and we don't have to worry about that. An additional benefit of lifting it up to the main wrapper is that we can now pass the recaptcha on any request should we need to.

src/base.ts
Averethel: I recall using it a couple of times when debugging. It's typed now so I'm not that much concerned here.

src/__tests__/base.test.ts
Averethel: I recall using cors modes for some debugging

src/base.ts
Averethel: We never use the word `Request`. In the context of the API client, we always talk about API calls so I liked @masone's suggestion

src/base.ts
Averethel: Jain. `ApiCallOptions` doesn't specify method and we don't want people setting method to `GET` when invoking `postData`

src/services/__tests__/dealer.test.ts
Averethel: Yeah thought about that too. Will do 👍

src/services/car/moneybackApplication.ts
Averethel: Why? So that consumer needs to know that `recaptchaToken` is passed in the second option object and `validateOnly` in the first one? That creates unnecessary complexity and makes usage cumbersome in my opinion.

src/types/params/index.ts
Averethel: It is exported internally, not externally. I've seen we typed the language in some calls and accepted strings in another, went with the concrete type.

lkappeler: Ah nvm messed that up as the file is called index....

src/base.ts
lkappeler: Ahh I did not ment to pass it form the consumer rather form the endpoint function as only a couple of endpoints will use it no strong opinion, up to you...

src/__tests__/base.test.ts
lkappeler: Ohh you just typed that internally. Messed that up... nvm

src/base.ts
lkappeler: yeah I see the pint on the method here but the rest should be consisten like if you want to have the generic fetch options for example (for debugging) I would expect to be able to set them in a post or put request the same way...

src/services/car/moneybackApplication.ts
lkappeler: I don't think it matters a lot at the end  so up to you...

My thinking behind would be to get it somewhat close to the request or fetch interface where the first argument is the resource and the second one the options (I realized body is an option there...). It's probably preference but thats like how I am used to interfaces of this kind... (It's kinda separation of concerns as the consumer could always pass the same request options interface no matter of the specific api request; I do see some benefits when it comes to maintanance as the generic request interface could live standalone. Given the very unlikely case that we want to add `validateOnly` that does somthing different than the option in this request for example )

src/base.ts
Averethel: Yeah, that's what I initially thought and then it ended up being a lot of endpoints that all manually transformed header so I lifted it.

src/base.ts
Averethel: That's why `RequestOptions` extends `ApiCallOptions` with the method only.
We have a `fetchPath` that's the lowest level wrapper around fetch and we have `postData` and fiends that are wrappers around `fetchPath`. We don't have a wrapper around the `get` request - is that what you're hinting at?

src/services/car/moneybackApplication.ts
Averethel: Seems like premature optimization to me... Let's cross the clashing options bridge when we get there ok?
I see your point but it does require the consumer to be aware of what is a custom option and what is a specific option and this means that consumer needs to know the implementation details which sounds a bit like smell to me...

src/base.ts
```diff
@@ -85,58 +40,78 @@ const getAuthorizationHeader = (accessToken = null) => {
...
-export interface RequestOptions {
-  isAuthorizedRequest?: boolean
+export const getHost = (host: string = null) => {
+  if (!apiClient.configuration.host) {
+    throw new Error(
+      'ApiClient not configured, please run: ApiClient.configure({ apiGatewayUrl: "your.api.gateway" }'
+    )
+  }
+
+  return host || apiClient.configuration.host
```
masone, line 50: what's the use case for passing a host to a function that is then returned transparently?

```
getHost("foo.local") // => foo.local
```

src/services/__tests__/catalogue.test.ts
masone: thanks, let's roll with it

masone (approved):
I gave this another scan, but not a deep look at every change. Looks like you were able to bundle some useful improvements, thanks for that

I'd ask for a full regression test for each project in one browser before moving on. I recommend only merging this PR once you validated the projects with a prerelease (to not block other peoples minor releases to be coupled to your breaking changes)

src/base.ts
Averethel: The idea was to extract getting the host based on options. It returns the host (either passed one or the configured one) but also guards that the API client is configured

src/apiClient.ts
```diff
@@ -1,11 +1,10 @@
...
-  dealerServiceUrl?: string
-  optionServiceUrl?: string
-  analyticsServiceUrl?: string
-  userNotificationServiceUrl?: string
+export interface ApiClientConfiguration {
+  host: string
+  debug?: boolean
+}
+
+interface ApiClientConfig {
```
lkappeler, line 6: Why do you need an additional internal type you you default the configuration to an empty object in the constructor?

src/base.ts
```diff
@@ -85,58 +40,78 @@ const getAuthorizationHeader = (accessToken = null) => {
   return { Authorization: `Bearer ${accessToken}` }
 }
 
-export interface RequestOptions {
-  isAuthorizedRequest?: boolean
+export const getHost = (host: string = null) => {
+  if (!apiClient.configuration.host) {
+    throw new Error(
+      'ApiClient not configured, please run: ApiClient.configure({ apiGatewayUrl: "your.api.gateway" }'
```
lkappeler, line 46:
```suggestion
      'ApiClient not configured, please run: ApiClient.configure({ host: "your.api.gateway" }'
```

src/base.ts
```diff
@@ -85,58 +40,78 @@ const getAuthorizationHeader = (accessToken = null) => {
...
+    ...headers,
+  }
+}
+
+export interface ApiCallOptions extends Omit<RequestInit, "method" | "body"> {
+  recaptchaToken?: string
   accessToken?: string
-  method?: "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
-  headers?: HeadersInit
+  headers?: Record<string, string>
```
lkappeler, line 71: Just a side node: fech accepts the headers in different variations than just an object, we can go with objects only for our use case...

src/base.ts
masone: that code still makes little sense to me even with your comment 🙈 
if it's just an internal thing, I'd roll with whatever you think is right. if it's external-facing, I'd get another opinion

src/types/models/autoAlarm.ts
```diff
@@ -7,22 +7,22 @@ export interface DealerSavedSearchQuery {
...
   bodyType: string[]
-  firstRegistrationYearFrom: number | null
-  firstRegistrationYearTo: number | null
-  mileageFrom: number | null
-  mileageTo: number | null
-  priceFrom: number | null
-  priceTo: number | null
-  horsePowerFrom: number | null
-  horsePowerTo: number | null
+  firstRegistrationYearFrom?: number
```
lkappeler, line 10: It was intentionally typed to be a number or null but not to be missing? I have in mind the application relies on this behavior as we run into issues with the form when the property was undefined. Also passing props form getServerSide props will have issues with undefined values as in `firstRegistrationYearFrom: undefined` will result in a serialization error... (U think it works fine if the property is not there...)

src/services/search/typeSearch.ts
```diff
@@ -75,14 +72,17 @@ export const fetchTypes = async (
...
-  query: SearchTypeQueryParams = {},
-  fields: string[] = [],
-  options: RequestOptions
-): Promise<WithValidationError<Facets>> => {
+export const fetchTypeFacets = async ({
+  query = {},
+  fields = [],
+  options = {},
+}: {
+  query?: SearchTypeQueryParams
```
lkappeler, line 80: Hmm interesting, I have in mind that I get linter errors in the project when marking properties that default to something as optional.

lkappeler (commented):
Please take a look at the comment on the changed auto alarm type. I think it changes the intended behavior and does not really fit into the interface change. (Might be off...)
Other than that some minor inputs, saw that you plan to update the docs 👍

src/apiClient.ts
Averethel: Because `host` is required in the external type and optional in the internal one.
As in you can't configure the `ApiClient` without the `host` you'll get the type error

src/services/search/typeSearch.ts
Averethel: What case are you talking about? That seems like an error in the linter rules TBH, can you check if that's still the case and create a ticket if so?

src/types/models/autoAlarm.ts
Averethel: hmm strange, I did not recall any form issues with listings where we use the optional types. I'll revert that change

lkappeler: The one with the serialization is probably in with the new next version. Don't know about the listings, perhaps the approach to default the values is inclined in the form component... (Or I messed up something 🙃)

src/services/search/typeSearch.ts
lkappeler: I might be of or it behaves differently when using it as part of an object..
Had that one in mind
image

src/apiClient.ts
lkappeler: Oh nvm messed that up 😂

src/types/models/autoAlarm.ts
Averethel: I did a quick test and I was able to interact with saved searches properly with those new types. Unless QA discovers something that I missed, let's keep it the way it is.

src/services/search/typeSearch.ts
Averethel: Okay, that's a different case.
In your example you have:
```
const foo = (param: string = "string") => ...
```
which is an argument of a function,
whereas here we have:
```
const foo = ({ bar = "string" }: { bar?: string })  => ....
```
which instantiates a field of the object.
It makes sense to error out in the first case since by assigning default you create two call signatures:
```
foo(param: string)
foo()
```
whereas in the second case it does not touch call signatures, only ensures that the field has a default value.

src/base.ts
Averethel: Yeah, everything here is internal. I'd rather have that separated from `fetch` wrapper since it makes things easier to grasp.

src/services/search/typeSearch.ts
lkappeler: Thanks for the explanation, thought somehow it would compare the call sinature of the argument the same way....

lkappeler (approved):

PR #181, opened 2020-10-21
feat: support searching for listing options

masone:
https://autoricardo.atlassian.net/browse/CAR-5391

extended the types to support querying for listing options, see https://autoricardo-ag.slack.com/archives/GBN03UZ3K/p1602594132293600


Averethel (approved):
With the `key` directive in the type, there's no need to explicitly whitelist every single search parameter.
Up to you whether you want to whitelist explicitly.

masone:
yeah, I saw that too... it seems that's what we're doing though, so I rolled with it :)

PR #182, opened 2020-10-23
feat: replace leadsAnalytics with leadsImpressions and add fetchLeads…

ufkaya:
BREAKING CHANGE

Ref:

> Endpoint GET /dealers/{dealerId}/leads/count should be used only for leads count. It returns exact number of message and call leads for the previous 30 days (sources are our DB and Matelso). Endpoint POST /dealers/{dealerId}/leads/impressions  is basically the same like the initial POST /dealers/{dealerId}/leads/analytics it returns leads stats for provided dimensions and function, we just removed time dimension from the request, we will always return data for the previous 30 days (source for this endpoint is GA).

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,9 +184,9 @@ Also accompanying modes and param types, as well as default values, are exported
 
 - [Analytics](https://carforyou-analytics-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics)
   - [fetchAnalyticsData](https://carforyou-analytics-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getListingsMetricsUsingPOST)
-  - [fetchLeadsImpressions](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsImpressionsUsingPOST)
+  - [fetchLeadsAnalytics](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsAnalyticsUsingPOST)
   - [fetchListingsAnalytics](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerListingsAnalyticsUsingPOST),
-  - [fetchLeadsCount](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsCountUsingGET)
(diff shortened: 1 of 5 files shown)


src/services/analytics.ts
```diff
@@ -23,23 +27,20 @@ export const fetchAnalyticsData = async ({
...
   dimensions,
-  query,
   options = {},
 }: {
   dealerId: number
   dimensions?: string[]
-  query?: { verificationDateFrom: string }
   options?: ApiCallOptions
 }): Promise<Array<CockpitAnalytics>> => {
   return postData({
```
Averethel, line 39: Is this still POST?

Averethel (commented):
Just to clarify - the last 30 days happen on the BE side now? What's the motivation for it? That's not the most flexible approach....

src/services/analytics.ts
ufkaya: yes because I send the function and dimensions

masone (approved):
I can't really speak to the functionality, but code looks just fine 👌

README.md
```diff
@@ -184,8 +184,9 @@ Also accompanying modes and param types, as well as default values, are exported
 
 - [Analytics](https://carforyou-analytics-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics)
   - [fetchAnalyticsData](https://carforyou-analytics-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getListingsMetricsUsingPOST)
-  - [fetchLeadsAnalytics](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsAnalyticsUsingPOST)
-  - [fetchListingsAnalytics](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerListingsAnalyticsUsingPOST)
+  - [fetchLeadsImpressions](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsImpressionsUsingPOST)
+  - [fetchListingsAnalytics](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerListingsAnalyticsUsingPOST),
+  - [fetchLeadsCount](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Analytics/getDealerLeadsCountUsingGET)
```
Averethel, line 189: Nitpicking but those live on reporting service, not analytics service.

Averethel (approved):

PR #184, opened 2020-10-28
feat: store purchase confirmations

masone:
context, see https://autoricardo.atlassian.net/browse/CAR-5251

preliminary information https://autoricardo-ag.slack.com/archives/C01A3ECLW7M/p1603874745000900?thread_ts=1603871702.000800&cid=C01A3ECLW7M

TODO:
- [ ] squash merge with semantic release commit message

This is my first contribution to the api client. Please give me a *thorough review* and sanity check for folder structure and file naming please.

## testing
there's a mock api in place currently, which behaves as follows:
- if the key ends with letter response is 200
- if the key ends with digit <= 5 response is 422, already added
- if the key ends with digit >5 response is 404 not found

http://localhost:3000/de/lead-feedback/782443/key-ok/reject
http://localhost:3000/de/lead-feedback/782443/key-already-added-1/reject
http://localhost:3000/de/lead-feedback/782443/key-not-found-9/reject


src/services/carSaleTracking/buyerFeedback.ts
```diff
@@ -0,0 +1,20 @@
+import { postData, ApiCallOptions } from "../../base"
+import { WithValidationError } from "../../types/withValidationError"
+
+export const addPurchaseConfirmation = async ({
```
lkappeler, line 4: To me the naming is a bit confusing as we add a feedback that either confirms or denies a purchase, might be jsut a personal view 🙃
It aligned with backed so it's fine...

src/services/carSaleTracking/buyerFeedback.ts
```diff
@@ -0,0 +1,20 @@
...
+  key,
+  confirmed,
+  options = {},
+}: {
+  key: string
+  confirmed: boolean
+  options?: ApiCallOptions
+}): Promise<WithValidationError> => {
+  return postData({
```
lkappeler, line 13: IIRC the post and put functions do not abstract the validation handling so you have to wrap the error handling in some try catches to fulfill the interface...
```
 try {
    const result = await putData({
      service: Service.SEARCH,
      path: `dealers/${dealerId}/listing-saved-searches/auto-alarms/${savedSearchId}`,
      body: savedSearch,
      options: { isAuthorizedRequest: true, ...options },
    })

    return {
      tag: "success",
      result,
    }
  } catch (error) {
    return handleValidationError(error, { swallowErrors: true })
  }
```
(Very happy to discuss the envelope approach in general but I think we should not mix patterns for now)

masone: I have been thinking about this as well, but as there is no returned data on this call (it's a fire and forget for us) and I have to rely on the response codes to react correctly in the application. As there is no validation required, I went with not wrapping them

Does that make sense? Otherwise I'm happy to discuss in a call! I was back and forth here too

src/services/carSaleTracking/buyerFeedback.ts
masone: https://car-sale-tracking-service.preprod.carforyou.ch/swagger-ui/index.html#/Buyer%20Feedback/addPurchaseConfirmationUsingPOST

the naming comes straight from the api call: 
```
/api/car-sale/buyer-feedback/key/{key}/add-purchase-confirmation
```

src/services/carSaleTracking/buyerFeedback.ts
masone: we decided to roll with it

lkappeler (approved):

PR #185, opened 2020-11-02
fix: cockpit model

ufkaya:


ralucacril (approved):

Averethel (changes requested):
Can we stick to the `key` directive here? I'd rather not whitelist every single thing we want to pass to the statistics call in the BE, this would mean we need to update the client every time we want a new dimension which isn't really feasible.

If the `toString` calls are annoying maybe:

```
export type CockpitAnalytics  = { count: number } & { [key: string]: string }
```
would work better?

ufkaya:
> Can we stick to the `key` directive here? I'd rather not whitelist every single thing we want to pass to the statistics call in the BE, this would mean we need to update the client every time we want a new dimension which isn't really feasible.
> 
> If the `toString` calls are annoying maybe:
> 
> ```
> export type CockpitAnalytics  = { count: number } & { [key: string]: string }
> ```
> 
> would work better?

no it is fine, then lets not change it

PR #187, opened 2020-11-10
feat: add matelso phone to dealer profile

ufkaya:
- [CAR-5392](https://autoricardo.atlassian.net/browse/CAR-5392)

### DIFF
--- a/src/services/dealer.ts
+++ b/src/services/dealer.ts
@@ -218,7 +218,7 @@ export const requestMatelsoIntegration = async ({
 
     return {
       tag: "success",
-      result: { ...result },
+      result,
     }
   } catch (error) {
     return handleValidationError(error, { swallowErrors: true })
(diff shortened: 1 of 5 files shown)


ralucacril (approved):

ralucacril (approved):

src/services/dealer.ts
```diff
@@ -198,3 +198,29 @@ export const putDealerPromotion = async ({
...
+        ...options,
+      },
+    })
+
+    return {
+      tag: "success",
+      result,
+    }
+  } catch (error) {
+    return handleValidationError(error, { swallowErrors: true })
```
lkappeler, line 224: What validation errors are we getting from that endpoint?

ufkaya: none, this is the default one that I use in case BE returns some errors, what should I use instead?

lkappeler: As far as I have in mind this is used to handle / format filed validation errors when sending data to the API for format the validation messages and since for this call I would the API expect to just succeed or reject so I don't see why we would need to add form validation handling here...

PR #190, opened 2020-11-18
allow overriding the api version per-call

masone:
ref https://autoricardo.atlassian.net/browse/CAR-5381

backend introduced a new version for the lead messages api. the api versioning applies per endpoint and we can't globally set the api version as we initially planned to. this adds capabilities to override the api version per-call

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -3,7 +3,7 @@ import { WithFieldStats, FieldsStats } from "./types/fieldStats"
 import { Facets } from "./types/facets"
 import { ValidationError } from "./types/withValidationError"
 
-import apiClient from "./apiClient"
+import apiClient, { ApiVersion } from "./apiClient"
 import { ResponseError } from "./responseError"
 import { SearchListing } from "./types/models/listing"
 import { WithTopListing } from "./types/topListing"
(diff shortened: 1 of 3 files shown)


src/base.ts
```diff
@@ -80,6 +82,7 @@ type Method = "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
 
 interface RequestOptions extends AuthorizedApiCallOptions {
   method?: Method
+  apiVersion?: ApiVersion
```
lkappeler, line 85: Don't you need to add it to the `ApiCallOptions` Interface so it's available/typed for external and POST/PUT/DELETE requests as well?

masone: indeed, just noticed that as well when working on the actual implementation

lkappeler: It's already there as you extend `AuthorizedApiCallOptions` also this is the internal interface so we probably wanna add it to the base interface `ApiCallOptions` instead...

masone: the applied change seems to do the trick

lkappeler: oke might have messed up something then :sli

lkappeler (approved):

src/base.ts
```diff
@@ -74,12 +76,14 @@ export interface ApiCallOptions extends Omit<RequestInit, "method" | "body"> {
...
 interface AuthorizedApiCallOptions extends ApiCallOptions {
   isAuthorizedRequest?: boolean
+  apiVersion?: ApiVersion
 }
 
 type Method = "GET" | "POST" | "PUT" | "PATCH" | "DELETE"
 
 interface RequestOptions extends AuthorizedApiCallOptions {
   method?: Method
+  apiVersion?: ApiVersion
```
Averethel, line 86: Do we need to add this twice? It should already be available here since you're extending `AuthorizedApiCallOptions`

src/base.ts
```diff
@@ -74,12 +76,14 @@ export interface ApiCallOptions extends Omit<RequestInit, "method" | "body"> {
 
 interface AuthorizedApiCallOptions extends ApiCallOptions {
   isAuthorizedRequest?: boolean
+  apiVersion?: ApiVersion
```
Averethel, line 79: This has nothing to with authorized API call, can you add it on base `ApiCallOptions` instead?

Averethel (changes requested):
Please check the interfaces there seems to be something off there

PR #191, opened 2020-11-18
Split first/last name in message lead

masone:
ref https://autoricardo.atlassian.net/browse/CAR-5381

backend introduced a new version for the lead messages api where the lead message name is split into first/last name. changes types and does the api v2 request

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -67,6 +67,8 @@ const buildHeaders = ({
   }
 }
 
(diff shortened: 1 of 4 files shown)


Averethel (approved):

lkappeler (approved):
Noice

GutierrezGo (approved):
After merging this change, we are able to split the name in the lead email templates, right?

masone:
correct. when someone uses the`name` field, they would need to use a version prior to this change. from this on forward, the names are split

PR #193, opened 2020-11-20
feat: add requestWhatsappIntegration api and whatsapp number to dealer profile

ufkaya:
- [CAR-5290 - Cockpit: Whatsapp CTA phone no. entry with confirmation + BO notification](https://autoricardo.atlassian.net/browse/CAR-5290)

### DIFF
--- a/src/services/dealer.ts
+++ b/src/services/dealer.ts
@@ -225,7 +225,7 @@ export const requestMatelsoIntegration = async ({
   }
 }
 
-export const requestWhatsappIntegration = async ({
+export const requestWhatsAppIntegration = async ({
   dealerId,
   whatsAppNumber,
   options = {},
--- a/src/services/dealer.ts
+++ b/src/services/dealer.ts
@@ -224,3 +224,31 @@ export const requestMatelsoIntegration = async ({
     return handleValidationError(error, { swallowErrors: true })
   }
 }
+
+export const requestWhatsappIntegration = async ({
+  dealerId,
+  whatsAppNumber,
+  options = {},
+}: {
+  dealerId: number
+  whatsAppNumber: string
+  options?: ApiCallOptions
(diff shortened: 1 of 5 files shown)


src/services/dealer.ts
```diff
@@ -224,3 +224,31 @@ export const requestMatelsoIntegration = async ({
...
+        ...options,
+      },
+    })
+
+    return {
+      tag: "success",
+      result,
+    }
+  } catch (error) {
+    return handleValidationError(error, { swallowErrors: true })
```
masone, line 252: can you help me understand - what errors do we swallow and why?
what validation errors could occur in this case?

masone (commented):
some questions on the error handling :)

src/services/dealer.ts
Averethel: If you add `{ swallowErrors: true }` response errors other than 400 and 422 (as well as other exceptions) will be returned as:
```
{
  tag: "error",
  message: "validation.other-error",
  errors: [],
}
```

that's how we typically handle non-critical API requests
(check the code for reference: https://github.com/carforyou/carforyou-api-client-pkg/blob/master/src/base.ts#L192)

Averethel (approved):

src/services/dealer.ts
ufkaya: 1. I used the same code (copy/paste) and I didnt check what that part does, thanks @Averethel 
2. we send the whatsappNumber, and we get the validation error if not populated

ralucacril (approved):

src/services/dealer.ts
masone: ah, so this is going to be used in a form... thanks for the explanation, makes sense

masone (approved):

PR #206, opened 2021-01-27
BREAKING CHANGE: new dealer listing search and count methods

Averethel:
`fetchDealerListings` and `fetchDealerListingsCount` have been moved to the new endpoints that use elastic search

Since we won't have archived listings on elastic due to the overwhelm the number of listings would create on indexes (which would slow the search down considerably as well as increase the costs by a lot) we also add separate endpoints for archived dealer listings (`fetchDealerArchivedListings` and `fetchDealerArchivedListingsCount`). Not that those endpoints are not filterable.
Both dealer and public search APIs now living on elastic allows us to simplify filter and sort parameter by a bunch.


Averethel:
@carforyou/frontend while I'd appreciate a review here I'm keeping it in draft until the APIs are not merged and released.

src/services/search/listingSearch.ts
```diff
@@ -39,27 +44,83 @@ export const fetchListingCount = async ({
...
+  sortType: ListingSortTypeParams.CREATED_DATE,
+  sortOrder: ListingSortOrderParams.DESC,
+}
+
+export const defaultUserPagination = {
   page: 0,
   size: 24,
 }
 
+export const defaultDealerPagination = {
```
lkappeler, line 103: What's the reason to have a different default pagination length for the User and the Dealer?

src/services/search/listingSearch.ts
```diff
@@ -92,6 +153,14 @@ const searchForListings = ({
   return postData({ path, body, options: otherOptions })
 }
 
+const sanitizeListing = ({
```
lkappeler, line 156: Whats is the function sanitizing? It looks more like converting some data based on the function called...
Btw why is the `decodeDate` function adding a month to the passed date this looks very strange to me...

src/services/search/__tests__/listingSearch.test.ts
```diff
@@ -70,6 +74,61 @@ describe("SEARCH service", () => {
...
+
+    it("unwraps count from json", async () => {
+      const count = 400
+      fetchMock.mockResponse(JSON.stringify({ count }))
+
+      const response = await fetchDealerListingsCount({
+        dealerId,
+        options: requestOptionsMock,
+      })
+      expect(response).toEqual(count)
```
lkappeler, line 91: Does this test make sense like that, I might miss something but isn't it just ensuring that fetch has been called in any way at some point in the stack...
Maybe it would make more sense to mock `fetchPath` in the base file and ensure that it was called with the correct props instead of testing if the fetch mock returns the mocked data...

lkappeler: Saw that the same may apply to other tests following the same pattern, no gonna comment there so we can have the discussion in one place

src/services/search/listingSearch.ts
lkappeler, line 103: Whats the reason so we need a different default pagination for the dealer than the user?

src/services/search/listingSearch.ts
```diff
@@ -128,11 +191,88 @@ export const fetchListings = async ({
...
+      ...options,
+    },
+    defaultSort: defaultDealerSort,
+    defaultPagination: defaultDealerPagination,
   })
 
   return sanitizeListingResponse(response)
 }
 
+const toCamelCase = (snakeCase) => {
```
lkappeler, line 224: Not sure if we should move that to lib code as someone might be more likely to look there in case the functionality is required somewhere else

src/services/search/listingSearch.ts
```diff
@@ -128,11 +191,88 @@ export const fetchListings = async ({
...
+
+  const sortOrDefault = {
+    sortType: sortType || defaultDealerSort.sortType,
+    sortOrder: sortOrder || defaultDealerSort.sortOrder,
+  }
+
+  const queryParams = {
+    page: pageOrDefault(page, defaultDealerPagination),
+    size: sizeOrDefault(size, defaultDealerPagination),
+    sort: `${toCamelCase(sortOrDefault.sortType)},${toCamelCase(
```
lkappeler, line 253: Why do we need to convert here would it be possible to type the param so we get the desired format?

src/services/search/listingSearch.ts
```diff
@@ -128,11 +191,88 @@ export const fetchListings = async ({
...
+      sortOrDefault.sortOrder
+    )}`,
+  }
+
+  const { content, ...response } = await fetchPath({
+    path: `dealers/${dealerId}/archived-listings${
```
lkappeler, line 259: You could also generate the query string and leave the key handling to the `toQueryString` as you should get an empty string if there are no keys.... And only check if we have a qs here

lkappeler (commented):
It looks good over all I am a bit concerned that a lot of the unit tests miss the actual functionality so one could potentially change the entire function (besides returning a fetch call) and the tests would still pass...

src/services/search/listingSearch.ts
Averethel: Gold listing

Averethel: Gold listing

src/services/search/listingSearch.ts
Averethel: We want to use the same enums for pagination for both ES (elastic search) and non-ES endpoints. They accept different formats.

src/services/search/__tests__/listingSearch.test.ts
Averethel: I mostly moved those around from the old implementation. It does ensure that the API response is processed the way we expect. We mainly mock raw fetch in the package that implements a lot of function on top of fetch to have a better overview if something breaks in the implementation of the wrappers.

src/services/search/listingSearch.ts
Averethel: It ensures that the date format is as expected (with a decoded date object). Kept the naming in line with the existing functions.
JS indexes months from `0` so yes, we do add one to have an indexing a sane person would expect when looking at a date object.

src/services/search/listingSearch.ts
Averethel: moved that to lib

src/services/search/listingSearch.ts
Averethel: Oh the check itself doesn't even make sense since there always are parameters (keys). Simplified.

ralucacril (approved):

src/lib/toCamelCase.ts
```diff
@@ -0,0 +1,11 @@
+const toCamelCase = (snakeCase) => {
```
lkappeler, line 1: It's not very likely to break I guess but a unit test would not hurt

src/lib/convertParams.ts
```diff
@@ -0,0 +1,23 @@
+import toCamelCase from "./toCamelCase"
+
+import { ListingSortParams, ListingSortTypeParams } from "../types/sort"
+
+// search APIs generally use elastic search, with the exception of the archived listing search
+// for the ease of use to not have a need for multiple data types we convert existing sort params
+// to the ones accepted by spring
+export const toSpringSortParams = (elasticSortParams: ListingSortParams) => {
```
lkappeler, line 8: thx for the comment 🙂

src/services/search/listingSearch.ts
lkappeler: Ah I see, js and dates 🙈

lkappeler (approved):
Thanks for the updates, looks good to me

src/lib/toCamelCase.ts
Averethel: thanks for the push

PR #209, opened 2021-01-28
feat: add warrantyText to the listing

ufkaya:
- [CAR-6047](https://autoricardo.atlassian.net/browse/CAR-6047)


ralucacril (approved):
you need to wait for BE to be ready, right?

ufkaya:
> you need to wait for BE to be ready, right?

Yes, they provided us the final names of the fields but they might change them during the implementation

PR #211, opened 2021-02-03
feat: added listPrice

ralucacril:
added new field


GutierrezGo (approved):

masone (approved):
good to finally having an english name for that field 👯‍♂️

PR #214, opened 2021-02-05
fix: allow sort by make and model

ralucacril:


Averethel (approved):
I'd go with `fix` here since the initial implementation was wrong

PR #216, opened 2021-02-05
fix: add type warrantyTypes

ufkaya:


src/types/models/listing.ts
```diff
@@ -77,6 +77,13 @@ interface Feature {
   startDate: string
 }
 
+export type WarrantyTypes =
+  | "from-date"
+  | "from-delivery"
+  | "from-first-registration"
+  | "none"
+  | "not-defined"
```
ufkaya, line 85: I will check with BE if we can remove `not-defined` and just use `null` instead

ralucacril (approved):

PR #217, opened 2021-02-10
feat: add globalErrors

ufkaya:
Reference https://autoricardo-ag.slack.com/archives/C8RRP2UQ4/p1612885405003400

### DIFF
--- a/src/services/__tests__/product.test.ts
+++ b/src/services/__tests__/product.test.ts
@@ -79,6 +79,7 @@ describe("Products service", () => {
         tag: "error",
         message,
         errors,
+        globalErrors: [],
       })
     })
   })
@@ -131,6 +132,7 @@ describe("Products service", () => {
         tag: "error",
         message,
         errors,
+        globalErrors: [],
       })
     })
   })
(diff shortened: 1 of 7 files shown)


src/services/__tests__/product.test.ts
```diff
@@ -79,6 +79,7 @@ describe("Products service", () => {
         tag: "error",
         message,
         errors,
+        globalErrors: [],
```
talamcol, line 82: I wonder if we should use the objectContaining matcher here and remove the globalError:
`
expect(response).toEqual({
       expect.objectContaining({
        tag: "error",
        message,
        errors
     }})
`
instead of adding it for every test. Because, if backend adds another field in the future, we have to touch all tests again, right? and we don't test the global error in this test here. 
But I'm not sure what's cleaner, up to you

talamcol (commented):
looks good!

src/services/__tests__/product.test.ts
ufkaya: Yes but I think is better for the future to be explicit, and I forgot to test the globalErrors for the warranty

masone (approved):
I'm not super familiar with api-client, but the changes seem sensible to me. If you want a more in-depth review, please choose someone else ;)

PR #219, opened 2021-02-12
New endpoint - Dealer Analytics

gzimh:
Exports new function `fetchDealerAnalytics` and its return type `DealerAnalytics` for the new (upcomming) endpoint `dealers/<dealerId>/analytics`


ufkaya (approved):
LGTM!
Please remove this file https://github.com/carforyou/carforyou-api-client-pkg/pull/219/files#diff-e2eb93a61ffd7877ea5c751abcb3a618e8e2e9a2073a27f66d4114fe10819f86
And update the readme with the new API

src/services/__tests__/analytics.test.ts
```diff
@@ -183,4 +186,83 @@ describe("Analytics service", () => {
...
+          {
+            name: "pdpViews",
+            value: 42,
+          },
+          {
+            name: "pdpViewsAvg",
+            value: 4,
+          },
+          {
+            name: "pdpViewsPerListingAvg",
```
disufi, line 208: I think this metric is not used at all, shall we correct it here?

src/services/__tests__/analytics.test.ts
```diff
@@ -183,4 +186,83 @@ describe("Analytics service", () => {
...
+          },
+        ],
+      },
+    ]
+
+    beforeEach(() => {
+      fetchMock.mockResponse(JSON.stringify(tilesData))
+    })
+
+    it("fetches tiles data", async () => {
```
disufi, line 231: we could also add a test case for the graph as well :)

disufi (approved):

Averethel (approved):

PR #222, opened 2021-02-15
feat: add vin

ufkaya:
Ref: [CAR-6038](https://autoricardo.atlassian.net/browse/CAR-6038)

### DIFF
--- a/src/services/__tests__/vehicle.test.ts
+++ b/src/services/__tests__/vehicle.test.ts
@@ -1,0 +1,65 @@
+import { fetchFrameNumberTypes } from "../vehicle"
+
+import { SearchType } from "../../lib/factories/type"
+
+const requestOptionsMock = { accessToken: "DUMMY" }
+
+describe("#fetchTypes", () => {
+  const types = SearchType()
+
+  beforeEach(() => {
+    fetchMock.mockResponse(
+      JSON.stringify({
+        frameNumber: "frameNumber",
+        productionYear: 2020,
+        types,
+      })
+    )
+  })
+
+  describe("when successful", () => {
+    it("returns types", async () => {
+      const response = await fetchFrameNumberTypes({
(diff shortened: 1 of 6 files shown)


masone (approved):

src/services/__tests__/vehicle.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    it("returns error messages", async () => {
+      const response = await fetchFrameNumberTypes({
+        query: { frameNumber: "", dealerId: 1316 },
+        options: requestOptionsMock,
+      })
+
+      expect(response.tag).toBe("error")
+
+      if (response.tag === "error") {
```
lkappeler, line 59: Why would you add a condition here? the expect above would fail otherwise...

src/services/vehicle.ts
```diff
@@ -0,0 +1,27 @@
...
+        isAuthorizedRequest: true,
+        ...options,
+      },
+    })
+    return {
+      tag: "success",
+      result: { content: result.types, pagination: null },
+    }
+  } catch (error) {
+    return handleValidationError(error)
```
lkappeler, line 25: How can a fetch request have a validation error/ whats the idea behind wrapping the fetch call into the envelope structure we use for POST and PUT requests?

src/services/__tests__/vehicle.test.ts
ufkaya: without the condition, I will get a type error, `Property 'errors' does not exist on type 'WithValidationError<Paginated<SearchType>>'`
We do the same thing in other tests

src/services/vehicle.ts
ufkaya: because `frameNumber` will be validated in BE

ralucacril (approved):

PR #223, opened 2021-02-17
feat: add carSales data builder

GutierrezGo:
References: 
> Jira https://autoricardo.atlassian.net/browse/CAR-6319

> API contract for fetching draft car sales https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getDealerCarSalesUsingGET

## Motivation and context
Extend the carSales api-client

## Thank you 🦙

### DIFF
--- a/src/types/models/carSales.ts
+++ b/src/types/models/carSales.ts
@@ -1,5 +1,4 @@
 import { ApiSearchListing } from "../../types/models/listing"
-import { SearchMessageLead } from "../../types/models"
 
 export interface Buyer {
   email: string
@@ -8,16 +7,24 @@ export interface Buyer {
   phone: string
 }
 
-export type FirstListingParams = Pick<
-  ApiSearchListing,
-  "id" | "firstRegistrationDate" | "price" | "mileage"
->
-
-export type SecondListingParams = SearchMessageLead["listing"]
+export interface CarSalesListing
+  extends Pick<
+    ApiSearchListing,
+    | "id"
+    | "firstRegistrationDate"
+    | "make"
+    | "model"
+    | "type"
+    | "price"
+    | "mileage"
+  > {
+  externalListingId: string
+  image: string
+}
 
 export interface CarSales {
   buyer: Buyer
   carSaleDate: string
   id: number
-  listing: [FirstListingParams, SecondListingParams]
+  listing: CarSalesListing
 }
--- a/src/types/models/carSales.ts
+++ b/src/types/models/carSales.ts
@@ -1,0 +1,25 @@
+export interface Buyer {
+  email: string
+  firstName: string
+  lastName: string
+  phone: string
+}
+
+export interface ListingQuery {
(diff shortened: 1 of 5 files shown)


src/types/models/carSales.ts
```diff
@@ -0,0 +1,25 @@
+export interface Buyer {
+  email: string
+  firstName: string
+  lastName: string
+  phone: string
+}
+
+export interface ListingQuery {
```
lkappeler, line 8: 1. Whats your motivation behind the naming?
2. It looks like a subset of the Listing interface instead of redefining the existing props we could use https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys to select the required sumbset

src/types/models/carSales.ts
```diff
@@ -0,0 +1,25 @@
...
+  id: number
+  image: string
+  make: string
+  mileage: number
+  model: string
+  price: number
+  type: string
+}
+
+export interface DealerCarSales {
```
lkappeler, line 20: Why the "Dealer" prefix?

lkappeler (commented):
Looks good so fat, a couple of thought on the types that where added

src/types/models/carSales.ts
GutierrezGo: I was thinking about the Dealer. But makes sense `CarSales`

src/types/models/carSales.ts
```diff
@@ -0,0 +1,23 @@
...
+import { SearchMessageLead } from "../../types/models"
+
+export interface Buyer {
+  email: string
+  firstName: string
+  lastName: string
+  phone: string
+}
+
+export type FirstListingParams = Pick<
```
lkappeler, line 11: - I think the fist second part approach is rather hard to understand as the consumer would have to custom combine the type.

What I had in mind was something like this
```
export interface CareSalesListing
  extends Pick<
    ApiSearchListing,
    | "id"
    | "firstRegistrationDate"
    | "price"
    | "mileage"
    | "externalListingId"
    | "make"
    | "model"
    | "type"
  > {
  image: string
}
```

src/types/models/carSales.ts
```diff
@@ -0,0 +1,23 @@
...
+  ApiSearchListing,
+  "id" | "firstRegistrationDate" | "price" | "mileage"
+>
+
+export type SecondListingParams = SearchMessageLead["listing"]
```
lkappeler, line 16: I would avoid extending a specific type for a view as it might change independently. The thought behind sharing was to reuse the pros of the listing we have defined. This one looks like a custom data structure for another feature (They could have used the same approach of extending the SearchListing there as well...)

GutierrezGo: I see, so I will unify it. Thanks

lkappeler (approved):
Thanks for the iteration looks very clean now

PR #224, opened 2021-02-17
feat: add support case

ralucacril:


ufkaya (approved):
LGTM

src/services/emailDelivery.ts
```diff
@@ -0,0 +1,32 @@
+import { WithValidationError } from "../types/withValidationError"
+import { ApiCallOptions, handleValidationError, postData } from "../base"
+
+export const supportCases = async ({
+  dealerId,
+  details,
```
Averethel, line 6: Is this the body of the message? How about calling it `message`?

src/services/emailDelivery.ts
```diff
@@ -0,0 +1,32 @@
+import { WithValidationError } from "../types/withValidationError"
+import { ApiCallOptions, handleValidationError, postData } from "../base"
+
+export const supportCases = async ({
```
Averethel, line 4: Let's go with:

```suggestion
export const sendSupportCase = async ({
```

1. we only send one case, singular reflects that intention
2. let's add `send` prefix to reflect that we're sending some data and be consistent with other naming

Averethel (approved):
The implementation looks good. I'd just slightly change the naming:

PR #225, opened 2021-02-18
feat: add callTracking endpoint

lkappeler:
Add the call tracking endpoint to the api client.

### DIFF
--- a/src/services/reporting/index.ts
+++ b/src/services/reporting/index.ts
@@ -1,6 +1,6 @@
-import { ApiCallOptions, ignoreServerSideErrors, postData } from "../../base"
-import { WithValidationError } from "../../index"
 import { createApiPathWithValidate } from "../../lib/path"
+import { WithValidationError } from "../../index"
+import { ApiCallOptions, ignoreServerSideErrors, postData } from "../../base"
 
 export const postTrackingData = async <T extends Record<string, unknown>>({
   listingId,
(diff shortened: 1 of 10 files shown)


src/types/models/tracking.ts
```diff
@@ -0,0 +1,8 @@
+import { Language } from "../../types/params"
+
+export type WhatsappEntry = Record<
+  "firstName" | "language" | "lastName" | "phone",
+  string
+>
```
masone, line 6: what's the reason to reach for a type and `Record` instead of using an interface as we usually do?

lkappeler: It's related to the dynamic type https://github.com/carforyou/carforyou-api-client-pkg/pull/225/files#diff-6d79f7439fbf4ceeb2d6e0ff0a243f47a09b9caf5acc63d081209fdac9b8ebd4R5 as you should not type it as `object` (see: [banned types](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md)) maybe I misinterpreted something here but II understood it as the recommended way with TS 2.0, the resulting type will be the same in this case but it matches the dynamic type passed to the extracted function

src/services/reporting/index.ts
```diff
@@ -0,0 +1,44 @@
+import { createApiPathWithValidate } from "../../lib/path"
+import { WithValidationError } from "../../index"
+import { ApiCallOptions, ignoreServerSideErrors, postData } from "../../base"
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const postTrackingData = async <T extends {}>({
```
lkappeler, line 6: I looked into accepting both types but this leads into other issues as typescript complains about the response to potentially be of the wrong type as the specific call only allows of one of the types in the response

masone (approved):
I feel at home with this! Thanks for the experiment with `Record` and sharing the learnings

PR #226, opened 2021-02-22
fix: added status in the query

GutierrezGo:
I forgot to add the status as a parameter in the query when I extended the car-sales data.

https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getDealerCarSalesUsingGET

Screenshot 2021-02-22 at 12 14 13

## Thank you 🔪

### DIFF
--- a/src/services/reporting/__tests__/carSales.test.ts
+++ b/src/services/reporting/__tests__/carSales.test.ts
@@ -46,6 +46,7 @@ describe("Car Sales", () => {
     it("unwraps content from json", async () => {
       const data = await fetchCarSales({
         dealerId: 123,
+        status: "confirmed",
         options: { accessToken: "DUMMY TOKEN" },
       })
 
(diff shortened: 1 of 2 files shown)


lkappeler:
@GutierrezGo Does that mean we need to make 2 requests to display the data in the overview, maybe it would make sense do discuss with backen if we can change the interface so we can pass multiple states to get the required data for the overview in one request?

lkappeler:
Ok we should be good here

> Each CarSale returned from API will have embedded rejection object. If it's null it means it's not rejected and opposite if it's not null it means CarSale is rejected. However this does not affect CarSale status, they will a be in DRAFT status, no matter is rejection=null or not. So you just need to send one singe request with query param status=draft and you need to handle this embedded rejection in order to display correct action on UI (reject/undo-reject)

lkappeler (approved):

PR #227, opened 2021-02-22
feat(carSales): create and delete rejections

lkappeler:
refs: https://autoricardo.atlassian.net/browse/CAR-6320

Adds PUT and DELETE actions for carSales rejections so the Dealer can reject draft carSales in the Client.

### DIFF
--- a/src/services/reporting/carSales.ts
+++ b/src/services/reporting/carSales.ts
@@ -2,7 +2,14 @@ import { Paginated } from "../../types/pagination"
 import { CarSales } from "../../types/models/carSales"
 
 import toQueryString from "../../lib/toQueryString"
-import { ApiCallOptions, fetchPath } from "../../base"
+import { WithValidationError } from "../../index"
+import {
+  ApiCallOptions,
+  deletePath,
+  fetchPath,
+  handleValidationError,
+  putData,
+} from "../../base"
 
 export const fetchCarSales = async ({
   dealerId,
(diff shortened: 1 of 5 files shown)


GutierrezGo (commented):
Not sure if we need the comment and the reason for the reject.

lkappeler:
> Not sure if we need the comment and the reason for the reject.

Yes they where just added, will update the PR

src/types/models/carSales.ts
```diff
@@ -28,3 +28,8 @@ export interface CarSales {
   id: number
   listing: CarSalesListing
 }
+
+export interface CarSaleRejection {
+  comment?: string
+  reason: "car-not-sold" | "wrong-buyer"
+}
```
GutierrezGo, line 35: I think we need to add `CarSalesRejection` into `CarSales` as `rejection: CarSalesRejection` or maybe I'm missing something there.

lkappeler: U right missed that it was missing in the data response as well, added it

src/services/reporting/carSales.ts
```diff
@@ -23,3 +30,71 @@ export const fetchCarSales = async ({
...
+  }
+}
+
+const getRejectionPath = ({
+  dealerId,
+  carSaleId,
+}: {
+  dealerId: number
+  carSaleId: number
+}) => `dealers/${dealerId}/care-sales/${carSaleId}/rejection`
```
GutierrezGo, line 55: I think is `car` intead `care`

GutierrezGo (approved):

src/services/reporting/carSales.ts
lkappeler: 🙈  - Good catch!

PR #236, opened 2021-03-02
feat: types and logic for lead messages created-date sorting

acapop77:
References CAR-6537

## Motivation and context
Handled needed API endpoint filter needed for `createdDate` sorting and introduced types and logic required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Also accompanying modes and param types, as well as default values, are exported
   - [`fetchTransmissionTypes`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Inventory%20Data/getTransmissionTypesUsingGET)
 - [Lead](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Lead)
   - [`sendMessageLead`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Lead/createMessageLeadUsingPOST)
+  - `fetchDealerMessageLeads`
 - [Leasing](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Leasing)
   - [`fetchLeasingFormUrl`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Leasing/generateLeasingProviderFormUrlUsingPOST)
   - [`sendLeasingInterest`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Leasing/createLeasingInterestUsingPOST)
--- a/src/lib/convertParams.ts
+++ b/src/lib/convertParams.ts
@@ -1,6 +1,9 @@
 import toCamelCase from "./toCamelCase"
 
 import {
+  LeadSortOrderParams,
+  LeadSortParams,
+  LeadSortTypeParams,
   ListingSortOrderParams,
   ListingSortParams,
   ListingSortTypeParams,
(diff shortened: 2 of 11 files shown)


src/lib/convertParams.ts
```diff
@@ -33,3 +36,25 @@ export const toSpringSortParams = (elasticSortParams: ListingSortParams) => {
...
+  const convertedSortOrder = toCamelCase(sortOrder)
+
+  switch (sortType) {
+    // We use alias in elastic, field name is createdDate
+    case LeadSortTypeParams.NEWEST:
+      return `createdDate,${toCamelCase(reverseSortLeadOrder(sortOrder))}`
+    default:
+      return `${toCamelCase(sortType)},${convertedSortOrder}`
+  }
+}
```
Averethel, line 60: Do we need to support both spring and elastic search here? The reason we have it for listings is that archived listings are handled with spring API whereas dealer and user listing searches are done via elastic and we want the parity between the two. Maybe we can simplify this if we only need to support spring?

src/types/sort.ts
```diff
@@ -29,3 +29,17 @@ export interface DealerSortParams {
...
 }
+
+export enum LeadSortTypeParams {
+  NEWEST = "NEWEST",
+}
+
+export enum LeadSortOrderParams {
+  ASC = "ASC",
+  DESC = "DESC",
+}
```
Averethel, line 40: Observation: there is always only `ASC` and `DESC` sort order. Do we have specialized types for other sorts as well? Would it make sense to ditch them and use a global one?

src/types/sort.ts
```diff
@@ -29,3 +29,17 @@ export interface DealerSortParams {
...
+
+export interface LeadSortParams {
```
Averethel, line 42: Observation: we keep creating sort param types that look very similar. Would it make sense to leverage a generic? Something along the lines of:

```typescript
interface SortParams<SortType> {
  sortOrder?: SortOrder
  sortType?: SortType
}
```

We can still export concrete types from the package if it's needed or let consumers build them on the fly

Averethel (commented):
Overall the changes look solid. I have a couple of questions/suggestions/observations, though. Let me know what do you think and how'd you like to proceed.

README.md
```diff
@@ -110,6 +110,7 @@ Also accompanying modes and param types, as well as default values, are exported
   - [`fetchTransmissionTypes`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Inventory%20Data/getTransmissionTypesUsingGET)
 - [Lead](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Lead)
   - [`sendMessageLead`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Lead/createMessageLeadUsingPOST)
+  - `fetchDealerMessageLeads`
```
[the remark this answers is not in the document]
acapop77: That is what I intended but there is none for this endpoint. I found for `/api/listings/{listingId}/message-leads` but there is none for `/api/dealers/{dealerId}/message-leads`.

src/lib/convertParams.ts
acapop77: I just went for using same names convention for filters where `NEWEST` being the one used for `createdDate` on listing so I just used same here for leads. And from that made all the changes just to make it work.

src/types/sort.ts
acapop77: Yes, we have exact same for Listing, is in the same file actually `ListingSortOrderParams`. They are same, only type names are different. I suppose if I make one default one for all and make needed changes, that way make them reusable as global one. Will make the changes.

src/types/sort.ts
acapop77: Yes, that makes sense. I will make the changes.

src/lib/convertParams.ts
acapop77: Removed spring helper functions part for leads. Won't even be using `NEWEST` as on leads we will use 'Date' name for sort and not 'Created' like we have on listing.

src/types/sort.ts
acapop77: Created generic `SortOrderParams` and switched lead na listing to use it, for listing made fallback export `SortOrderParams as ListingSortOrderParams` just as a fail-safe in case is used on other parts/repos of the project.

src/types/sort.ts
acapop77: I've made generic type just as suggested with a slight change as we have alternate `SortTypes` (ListingSortTypeParams|LeadSortTypeParams) so it takes a type variable:
```
interface SortParams<T> {
  sortOrder?: SortOrder
  sortType?: T
}
```
Also made fail-safe/fallback export `SortParams as ListingSortParams`. This new `ListingSortParams` now needs sortType implied and as such will need to make that change where is used. Or maybe I should still keep exiting `ListingSortParams` that does not require `sortType` ?

README.md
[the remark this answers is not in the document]
Averethel: Hmm, that's strange. It should be generated based on the code. Maybe we're looking in the wrong place or there's something else going on. CC @kvatrom @lukicsasa

Averethel (approved):

README.md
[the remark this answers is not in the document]
acapop77: Yep, only in [Buyer Service](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Money%20Back%20Guarantee%20Application) section we have leads mentioned.

PR #237, opened 2021-03-03
feat: bulk transfer to manual

Averethel:
References CAR-6668

## Motivation and context

Adds new API call for bulk transfer to manual

### DIFF
--- a/src/services/car/inventory.ts
+++ b/src/services/car/inventory.ts
@@ -286,6 +286,36 @@ export const transferDealerListingToManual = async ({
   }
 }
 
+export const transferDealerListingsToManual = async ({
+  dealerId,
+  listingIds,
+  options = {},
+}: {
+  dealerId: number
+  listingIds: number[]
+  options: ApiCallOptions
+}): Promise<WithValidationError> => {
+  try {
+    await postData({
+      path: `/api/dealers/${dealerId}/listings/transfer-to-manual`,
+      body: {
+        elements: listingIds,
+      },
+      options: {
+        isAuthorizedRequest: true,
+        ...options,
+      },
+    })
+  } catch (error) {
+    return handleValidationError(error)
+  }
+
+  return {
+    tag: "success",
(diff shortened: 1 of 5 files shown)


ralucacril (approved):

src/services/car/inventory.ts
```diff
@@ -286,6 +286,36 @@ export const transferDealerListingToManual = async ({
...
+      options: {
+        isAuthorizedRequest: true,
+        ...options,
+      },
+    })
+  } catch (error) {
+    return handleValidationError(error)
+  }
+
+  return {
```
lkappeler, line 313: Wont this potentially return before the promise is fullfilled so in case of an error we might get the wrong return value?

Averethel: No, there's an `await` in line [209](https://github.com/carforyou/carforyou-api-client-pkg/pull/237/files#diff-07e17215dafd3af97dcd5add6412634ccd2600a5bd5f046a56a8ffa81fef73c6R299)

lkappeler: yes but the await is inside the try catch block, https://playcode.io/745138/

Averethel: `setTimeout` behaves differently than promises. Also in the linked playground, you're not awaiting for the timed out function to finish execution (so it's similar to an example that would omit `await`).
Here's the example that works closer to what you want to test: https://playcode.io/745139/

It's the pattern we use already in the project and we have good test coverage here, I'm sure we would have caught an issue if it was one.

lkappeler: nvm, I messed something up in my head, thanks for checking

lkappeler (approved):

src/services/car/inventory.ts
Averethel: No worries. it's good to check things and ask these questions. There was a risk that something was not right there 👍

PR #241, opened 2021-03-04
feat: adds dealer-default-listing apis

ralucacril:
https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Dealer%20Default%20Listing

## Motivation and context

Adds the endpoints for dealer default listing data


src/types/models/listing.ts
```diff
@@ -19,6 +19,30 @@ export interface DealerListingImages {
...
+  warrantyMileage: number
+  warrantyStartDate: string
+  warrantyType: string
+}
+
+export interface DealerDefaultListing
+  extends DealerDefaultAdditionalServices,
+    DealerDefaultDescription,
+    DealerDefaultWarranty {
+  description: string
```
Averethel, line 42: wouldn't `DealerDefaultDescription` cover that already?

src/services/car/defaultListing.ts
```diff
@@ -0,0 +1,97 @@
...
+  DealerDefaultWarranty,
+} from "../../types/models/listing"
+import {
+  ApiCallOptions,
+  fetchPath,
+  handleValidationError,
+  putData,
+} from "../../base"
+
+export const fetchDealerDefaultListing = async ({
```
Averethel, line 14: I know we name our function after backend paths, but I'd expect it to return `Listing` when I read the name.
How about:
```suggestion
export const fetchDealerDefaultListingData = async ({
```

Averethel (commented):
Solid starting point. I have some feedback on the naming/types, but other than it looks really good.

src/types/models/listing.ts
ralucacril: i removed the DealerDefaultDescription but it doesnt appear in github. maybe they have issues.

i did ```
export interface DealerDefaultListing
  extends DealerDefaultAdditionalServices,
    DealerDefaultWarranty {
  description: string
}
```

File: src/services/car/defaultListing.ts [hunk 3]
```diff
@@ -0,0 +1,97 @@
...
+
+export const saveDealerDefaultDescription = async ({
+  dealerId,
+  description,
+  options = {},
+}: {
+  dealerId: number
+  description: string
+  options?: ApiCallOptions
+}): Promise<WithValidationError<{ description: string }>> => {
```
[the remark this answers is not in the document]
ralucacril: i removed it. github doesnt get my new changes.

src/types/models/listing.ts
Averethel: Oh, you probably did that mid-review. I think it makes sense to keep it for the sake of consistency.

Averethel (approved):

PR #244, opened 2021-03-08
feat: add setImage api

ufkaya:
References [CAR-6437](https://autoricardo.atlassian.net/browse/CAR-6437)

### DIFF
--- a/src/services/dealer.ts
+++ b/src/services/dealer.ts
@@ -253,3 +253,31 @@ export const setImage = async ({
     return handleValidationError(error, { swallowErrors: true })
   }
 }
+
+export const putDealerDescription = async ({
+  dealerId,
+  description,
+  options = {},
+}: {
+  dealerId: number
+  description: string
+  options?: ApiCallOptions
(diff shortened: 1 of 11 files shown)


GutierrezGo (approved):

src/services/dealer.ts
```diff
@@ -225,3 +225,31 @@ export const setLogo = async ({
...
   }
 }
+
+export const setImage = async ({
+  dealerId,
+  image,
+  options = {},
+}: {
+  dealerId: number
+  image: string
```
Averethel, line 235: Can we call it `imageUrl` or `imageS3Key`? When I read image here I expect an image object.

masone: keep it in sync with the other places where we upload images :)

ufkaya: what do you mean? Lets keep `image` since this is the field name in BE

masone: yeah, I guess that's what I mean... I'd value consistency with other api calls and backends naming over clarity

PR #246, opened 2021-03-09
feat: adds fields to use for global data on listing level

ralucacril:
## Motivation and context

Expose new fields for globalData

### DIFF
--- a/src/lib/factories/listing.ts
+++ b/src/lib/factories/listing.ts
@@ -126,6 +126,9 @@ const defaults: ListingType = {
   buyNowInProgress: false,
   transferredToManual: false,
   hidden: false,
+  useDefaultAdditionalServices: false,
+  useDefaultDescription: false,
+  useDefaultWarranty: false,
 }
 
 export function Listing(attributes = {}): ListingType {
@@ -266,6 +269,9 @@ export function EmptyListing(): ListingType {
     fullTankIncluded: undefined,
     expertInstructionIncluded: undefined,
     otherServices: undefined,
+    useDefaultAdditionalServices: undefined,
+    useDefaultDescription: undefined,
+    useDefaultWarranty: undefined,
   }
 }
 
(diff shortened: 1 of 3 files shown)


Averethel (approved):

GutierrezGo (approved):

src/lib/factories/listing.ts
```diff
@@ -266,6 +269,9 @@ export function EmptyListing(): ListingType {
     fullTankIncluded: undefined,
     expertInstructionIncluded: undefined,
     otherServices: undefined,
+    useDefaultAdditionalServices: undefined,
+    useDefaultDescription: undefined,
+    useDefaultWarranty: undefined,
```
masone, line 274: wouldn't that contradict the type definition below?

ralucacril: sorry, what?

masone: I might be misinterpreting things here, but the types say:
```
  useDefaultAdditionalServices: boolean
  useDefaultDescription: boolean
  useDefaultWarranty: boolean
```

and therefore the value can't be null/undefined. no?

ralucacril: check this file https://github.com/carforyou/carforyou-api-client-pkg/blob/master/src/lib/factories/listing.ts from MASTER. is about emptyListing.

masone: I don't know what that , but as you're confident, let's roll!

Averethel: @masone in typescript `undefined` and `null` are so-called bottom type. That means they are a bit special - they belong to every type that exist.

You can write:
```typescript
const foo: boolean = undefined
```
and you won't get a type error.

The concept behind this is so that you can handle not-initialized variables or empty values with ease, which is exactly what we're doing here.

lkappeler: Sidenote: thats what we plan to disable in the future https://autoricardo.atlassian.net/browse/CAR-5843 (

PR #247, opened 2021-03-09
fix: api call for global Data

ralucacril:
in a previous release i nested wring the params


Averethel (approved):

Averethel:
I'd go with `fix` instead of a `feat` here

PR #256, opened 2021-03-17
feat: add search api to leads

djanov:
References [CAR-5982](https://autoricardo.atlassian.net/browse/CAR-5982)

## Motivation and context
add for search api for the leads

### DIFF
--- a/src/types/params/leads.ts
+++ b/src/types/params/leads.ts
@@ -1,7 +1,7 @@
 import { LeadSortTypeParams, SortParams } from "../../types/sort"
 
 import { PaginationParams } from "./index"
-
 export interface LeadQueryParams extends PaginationParams {
   sort: SortParams<LeadSortTypeParams>
+  searchQuery: string
 }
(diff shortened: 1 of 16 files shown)


src/types/params/types.ts
```diff
@@ -21,6 +21,9 @@ export interface TypeFiltersParams {
   productionYear?: number
   imagesCountGroup?: string[]
 }
+export interface LeadTypeFiltersParams {
+  searchQuery?: string
```
masone, line 25: what is this required for? wouldn't one use `LeadQueryParams` instead? this type is not used anywhere, but only exported

djanov: I added that like the `TypeFiltersParams` and use it in DH, but I just checked I don't really use it for the search but for the `filters `and `derivedFilters` in the params.ts, but we will need this later, but for now I will delete here and on DH too.

src/types/params/leads.ts
```diff
@@ -1,7 +1,7 @@
 import { LeadSortTypeParams, SortParams } from "../../types/sort"
 
 import { PaginationParams } from "./index"
-
 export interface LeadQueryParams extends PaginationParams {
   sort: SortParams<LeadSortTypeParams>
+  searchQuery: string
```
lkappeler, line 6: Should the param be mandatory? Look like your are defaulting it in the search function...

djanov: @lkappeler you are right, it should be optional, my bad 🤦‍♂️

lkappeler (approved):

PR #258, opened 2021-03-19
feat: adds new method for bulk purchasing

gzimh:
References CAR-6428


gzimh:
@masone wasnt sure in what section of README.MD to document this new method. The existing (singular) one wasn't documented.

masone (approved):
the entry in the readme would be appreciated. the structure is trying to mimic how backend organizes the APIs. you should be able to figure it out looking at the swagger docs

disufi (approved):

PR #259, opened 2021-03-19
feat: add call leads API endpoint logic handle, create new types, tes…

acapop77:
feat: call leads API endpoint

References CAR-6636

## Motivation and context

Added Call Leads API endpoint logic

### DIFF
--- a/src/types/models/index.ts
+++ b/src/types/models/index.ts
@@ -166,6 +166,25 @@ export interface SearchMessageLead extends MessageLead {
   createdDate: string
 }
 
+export interface SearchCallLead {
+  id: number
+  listingId: number
+  listing: {
+    externalListingId: string
+    firstRegistrationDate: string
+    id: number
+    image: string
+    make: string
+    mileage: number
+    model: string
+    price: number
+    type: string
+  }
+  callerNumber: string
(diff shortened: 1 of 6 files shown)


src/types/models/index.ts
```diff
@@ -166,6 +166,25 @@ export interface SearchMessageLead extends MessageLead {
...
+    externalListingId: string
+    firstRegistrationDate: string
+    id: number
+    image: string
+    make: string
+    mileage: number
+    model: string
+    price: number
+    type: string
+  }
```
Averethel, line 182: is this exactly the same as for `SearchMessageLead`? Can it be extracted to a shared interface so it's easier to align changes between the two?

Averethel (approved):

src/types/models/index.ts
acapop77: Good point. There is two that are not same but at the same time, even do they are in the API docs in swagger we don't actually get that data:
```
    firstRegistrationDate: string
    id: number
```
I'll extract and make those two optional.

acapop77: Changed it. Extracted lead listing par to `MessageLeadListing` interface.

PR #262, opened 2021-03-19
fix: never return undefined for date fields

masone:
References https://autoricardo.atlassian.net/browse/CAR-5595

## Motivation and context

In listings, we get errors like the following and started cleaning `undefined` values with helpers. turns out this happens only for specific date fields, only on specific api endpoints:

```
SerializableError: Error serializing `.listing.firstRegistrationDate` returned from `getServerSideProps` in "/[languag
```

The solution is to return an explicit `null` and nextjs over in the projects is happy


lkappeler (approved):
NIce thanks

ralucacril (approved):

PR #265, opened 2021-03-22
fix: added authorized request

GutierrezGo:
References https://autoricardo.atlassian.net/browse/CAR-6693

## Motivation and context
We need to add the accessToken for the authorized request.


lkappeler (approved):
👍

PR #272, opened 2021-04-14
feat: add new verified flag to the listing type

talamcol:
References [CAR-7147](https://autoricardo.atlassian.net/browse/CAR-7147)

## Motivation and context

New field for verified on the SRP and PDP (backend is not yet ready)

image


Averethel (approved):

GutierrezGo (approved):
It looks good to me

PR #274, opened 2021-04-15
feat: add buyerFeedbackBatch data builder

GutierrezGo:
References https://autoricardo.atlassian.net/browse/CAR-6765

API ->  https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getBuyerFeedbackBatchUsingGET

## Motivation and context

Extend the buyer feedback batch api-client

## Thank you 🐙

### DIFF
--- a/src/types/models/buyerFeedbackBatch.ts
+++ b/src/types/models/buyerFeedbackBatch.ts
@@ -1,0 +1,23 @@
+import { ApiSearchListing } from "../../types/models/listing"
+
+export interface BuyerFeedbackDealer {
+  id: number
+  name: string
+  location: {
+    city: string
+    zipCode: string
+  }
+}
+
+export interface BuyerFeedbackListing
+  extends Pick<ApiSearchListing, "id" | "make" | "model" | "price" | "type"> {
+  image: {
+    s3Key: string
+  }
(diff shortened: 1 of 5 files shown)


ufkaya (approved):
LGTM

src/types/models/buyerFeedbackBatch.ts
```diff
@@ -0,0 +1,23 @@
...
+  id: number
+  name: string
+  location: {
+    city: string
+    zipCode: string
+  }
+}
+
+export interface BuyerFeedbackListing
+  extends Pick<ApiSearchListing, "id" | "make" | "model" | "price" | "type"> {
```
masone, line 13: can you quickly explains what this TS magic does?

GutierrezGo: With this utility `Pick<Type, Keys>` we pass an existing Type and set the keys that we need from the Type. Therefore, there's no need to repeat again the properties that we have already declared in another Type.

masone: got it. so we end up exposing a Listing-ish but with a limited amount of fields

masone (approved):

PR #283, opened 2021-04-29
fix: ensure location filter is converted when getting count

Averethel:
References [CAR-7301](https://autoricardo.atlassian.net/browse/CAR-7301)

## Motivation and context
When getting listings count we were not merging `cityId` and `radius` into a location object, which resulted in count not being updated when applying the radius filter


masone (approved):
ouch. the price of our past decisions :)

PR #284, opened 2021-04-29
feat: add vehicleUniverse to the listing

ufkaya:
## Motivation and context

BE is adding the vehicleUniverse on their side but we will change it and we will send it when we create a new listing


Marine-Berthier (approved):

masone:
what is this attribute for?

ufkaya:
> what is this attribute for?

we define the type of the vehicle which it can be car utility ..., for now, is always car

PR #285, opened 2021-05-04
feat: add bulkDealerFeedback

Marine-Berthier:
References [CAR-7202](https://autoricardo.atlassian.net/browse/CAR-7202?atlOrigin=eyJpIjoiOTU4ZmUyNDAxOWQ2NGM4YWJlMDU5NjU0NWU3YjI0YmIiLCJwIjoiaiJ9)
https://car-sale-tracking-service.preprod.carforyou.ch/swagger-ui/index.html#/Dealer%20Feedback/bulkCreateDealerFeedbackUsingPOST

## Motivation and context

Add new api to store bulk data collected by the dealer feedback.

Thank you 😃

### DIFF
--- a/src/services/carSaleTracking/bulkDealerFeedback.ts
+++ b/src/services/carSaleTracking/bulkDealerFeedback.ts
@@ -1,4 +1,5 @@
-import { ApiCallOptions, postData } from "../../base"
+import { WithValidationError } from "../../types/withValidationError"
+import { ApiCallOptions, handleValidationError, postData } from "../../base"
 
 export const postBulkDealerFeedback = async ({
   dealerId,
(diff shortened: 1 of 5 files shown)


src/services/carSaleTracking/bulkDealerFeedback.ts
```diff
@@ -0,0 +1,22 @@
...
+    purchaseConfirmed: boolean
+  }[]
+  options?: ApiCallOptions
+}): Promise<void> => {
+  return postData({
+    path: `dealers/${dealerId}/listings/car-sale/dealer-feedback`,
+    body: {
+      elements,
+    },
+    options,
```
GutierrezGo, line 20: I think in options you need the authorized Request.

ufkaya (approved):
LGTM

src/services/carSaleTracking/bulkDealerFeedback.ts
lkappeler: IIRC it's only one of the purposes of the options they are also used to set custom fetch options.
Regarding auth I would assume that the request needs authentication as the single feedback one does @Marine-Berthier can you check that with backend and add it if needed

Marine-Berthier: Yes, will do !

masone (approved):
nice one! high five! ✋

PR #288, opened 2021-05-12
Add "other" option for car sale rejection

Marine-Berthier:
References [CAR-7310](https://autoricardo.atlassian.net/browse/CAR-7310?atlOrigin=eyJpIjoiYTA1NTgxMWQ2YzdmNGIyODgzY2FhNWRkZDc1MDMyY2EiLCJwIjoiaiJ9)

## Motivation and context

In additon to "wrong-buyer" and "car-not-sold" we want to add "other"


src/types/models/carSales.ts
```diff
@@ -9,7 +9,7 @@ export interface Buyer {
 
 export interface CarSaleRejection {
   comment?: string
-  reason: "car-not-sold" | "wrong-buyer"
+  reason: "car-not-sold" | "wrong-buyer" | "other"
```
Marine-Berthier, line 12: I am not sure what more I need to do here. 
Nothing will change except that `comment` will become mandatory in case `reason` is "other". 
Is it something that should be handled here or in DH ?

masone: I'm not crazy good at this, but there is a way to define the type in a way that only when it's `other`, it has to have a `comment` and all other reasons must not have it. I think the concept is https://basarat.gitbook.io/typescript/type-system/discriminated-unions and it's something along those lines:

```
type CarSaleRejectionChoice = {
  reason: "car-not-sold" | "wrong-buyer"
};
type CarSaleRejectionOther = {
  reason: "other,
  comment: string
};
type CarSaleRejection = CarSaleRejectionChoice | CarSaleRejectionOther;
```

I'm not sure what implications it has on the app. depending how the form is returning the data, it might involve manually typing the data for each case. want to give it a try anyway?

if it causes too much of a hassle, I'd run with the type you went for. the api will validate the data too

masone (approved):

src/types/models/carSales.ts
lkappeler: I think it's fine to keep the comment as optional in the type like you did, the API will return a validation error that covers the business logic in case the reason is "other" and we send an empty comment.

lkappeler (approved):

src/types/models/carSales.ts
Marine-Berthier: Thank you Lorenzo and Christian for your feedback !

PR #290, opened 2021-05-17
feat: added carSalesCount data

GutierrezGo:
References https://www.figma.com/file/p8RIdxT0mcEQRf1g2dzpHa/2021---Q1---Reporting-%26-Billing?node-id=1207%3A295

Sales vs Capping + current transactions - https://car-sale-tracking-service.preprod.carforyou.ch/swagger-ui/index.html#/Car%20Sale/countCarSalesUsingGET.

## Motivation and context

Extend api-client for the widgets

## Thank you 🦀

### DIFF
--- a/src/services/reporting/carSales.ts
+++ b/src/services/reporting/carSales.ts
@@ -1,5 +1,5 @@
 import { Paginated } from "../../types/pagination"
-import { CarSales } from "../../types/models/carSales"
+import { CarSaleCount, CarSales } from "../../types/models/carSales"
 
 import toQueryString from "../../lib/toQueryString"
 import { CarSaleRejection, WithValidationError } from "../../index"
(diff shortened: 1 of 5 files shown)


src/services/reporting/carSales.ts
```diff
@@ -98,3 +98,21 @@ export const deleteCarSalesRejection = async ({
...
+  status?: string
+  hasRejection?: boolean
+  options?: ApiCallOptions
+}): Promise<CarSaleCount> => {
+  const query = toQueryString({ status, hasRejection })
+  return fetchPath({
+    path: `dealers/${dealerId}/car-sales/count${query ? `?${query}` : ""}`,
+    options: { isAuthorizedRequest: true, ...options },
+  })
+}
```
Averethel, line 118: How about returning a number directly?
```suggestion
}): Promise<number> => {
  const query = toQueryString({ status, hasRejection })
  const { count } =  fetchPath({
    path: `dealers/${dealerId}/car-sales/count${query ? `?${query}` : ""}`,
    options: { isAuthorizedRequest: true, ...options },
  })
  return count
}
```

src/types/models/carSales.ts
```diff
@@ -1,5 +1,9 @@
 import { ApiSearchListing } from "../../types/models/listing"
 
+export interface CarSaleCount {
+  count: number
+}
+
```
Averethel, line 6: This is not needed anymore, now that we switched to number.

Averethel (approved):

src/types/models/carSales.ts
GutierrezGo: True! thanks

PR #301, opened 2021-06-02
CAR-7534: Remove deprecated publish/validate endpoint

talamcol:
References [CAR-7534](https://autoricardo.atlassian.net/browse/CAR-7534)

## Motivation and context

The endpoint listings/validate is now doing the same as publish/validate did before. As a result, we can remove this endpoint here.

Related DH PR: https://github.com/carforyou/carforyou-dealerhub-web/pull/1314


acapop77 (approved):
Solid! 🙌

lkappeler (approved):

PR #302, opened 2021-06-11
chore: update node to v14

masone:
following our convention of using the lts release


lkappeler (approved):
Hmm looks like that's where the fetch mock warnings in DH originate...

PR #308, opened 2021-06-21
feat: fix carSalesRejection

Marine-Berthier:
Related to [this PR](https://github.com/carforyou/carforyou-dealerhub-web/pull/1357)

### DIFF
--- a/src/services/reporting/carSales.ts
+++ b/src/services/reporting/carSales.ts
@@ -34,12 +34,12 @@ export const fetchCarSales = async ({
 const envelopeRequest = async (handler): Promise<WithValidationError<null>> => {
   try {
     await handler
+    return {
+      tag: "success",
+      result: null,
+    }
   } catch (error) {
     return handleValidationError(error)
-  }
-  return {
-    tag: "success",
-    result: null,
   }
 }
 
--- a/src/services/reporting/carSales.ts
+++ b/src/services/reporting/carSales.ts
@@ -39,7 +39,7 @@ const envelopeRequest = async (handler): Promise<WithValidationError<null>> => {
       result: null,
     }
   } catch (error) {
-    return handleValidationError(error)
+    return handleValidationError(error, { swallowErrors: true })
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/services/reporting/carSales.ts
```diff
@@ -34,15 +34,12 @@ export const fetchCarSales = async ({
...
   } catch (error) {
-    return handleValidationError({
-      error,
-      options: { swallowErrors: true },
-    })
+    return handleValidationError(error)
+  }
+  return {
+    tag: "success",
+    result: null,
```
ufkaya, line 42: Why did you move it outside the `try()`?

Marine-Berthier: Because I tried to stick to the way we usually try / catch, but it can remain inside actually 👍

src/services/reporting/carSales.ts
```diff
@@ -39,10 +39,7 @@ const envelopeRequest = async (handler): Promise<WithValidationError<null>> => {
       result: null,
     }
   } catch (error) {
-    return handleValidationError({
-      error,
-      options: { swallowErrors: true },
-    })
+    return handleValidationError(error)
```
ufkaya, line 42: Are you sure that it is related to the `swallowErrors` option? How it is related?

Marine-Berthier: When `swallowErrors` is true then the response is undefined. 
But to be honest I don't understand why. Here : https://github.com/carforyou/carforyou-api-client-pkg/blob/2de06f89567613bb5036ccf8090b0d01c9154908/src/base.ts#L207, our `error.name` is "ResponseError" so it should not matter

src/services/reporting/carSales.ts
```diff
@@ -39,10 +39,7 @@ const envelopeRequest = async (handler): Promise<WithValidationError<null>> => {
       result: null,
     }
   } catch (error) {
-    return handleValidationError({
-      error,
-      options: { swallowErrors: true },
-    })
+    return handleValidationError(error, { swallowErrors: true })
```
Marine-Berthier, line 42: @ufkaya I think I found the "real mistake". Before it was `handleValidationError({error, options :{ swallowErrors: true }})`
This would explain why e`rror.statusCode` was not enough and that I had to `error.error.statusCode`

ufkaya: Nice! Good catch!

ufkaya (approved):
👏

PR #313, opened 2021-07-07
fix: fetchFrameNumberTypes return empty array if no result

ufkaya:
References [CAR-7875](https://autoricardo.atlassian.net/browse/CAR-7875)

## Motivation and context

https://autoricardo-ag.slack.com/archives/C8RRP2UQ4/p1624284136079200

When I search for VIN = WB1043001AZW29580 on Dealer Hub the list of results has 4132 pages because it returns an empty list of type ids which we use to fetch the types

## Test
https://github.com/carforyou/carforyou-dealerhub-web/pull/1389

### DIFF
--- a/src/services/vehicle.ts
+++ b/src/services/vehicle.ts
@@ -25,6 +25,16 @@ export const fetchFrameNumberTypes = async ({
       },
     })
 
+    if (!typeIds.length) {
+      return {
+        tag: "success",
+        result: {
+          content: [],
+          pagination: null,
+        },
+      }
+    }
+
     const types = await fetchTypes({
       query: { id: typeIds.map(({ id }) => id), page, size },
       options,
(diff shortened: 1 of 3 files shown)


Averethel (approved):
Can you add a test for this?

src/services/vehicle.ts
```diff
@@ -25,6 +25,16 @@ export const fetchFrameNumberTypes = async ({
...
+      return {
+        tag: "success",
+        result: {
+          content: [],
+          pagination: null,
+        },
+      }
+    }
+
     const types = await fetchTypes({
```
lkappeler, line 38: Is `fetchTypes` really supposed to return all types if we query for an empty list of ids?

ufkaya: Yes, this is how BE implemented it, check this slack discussion for more details https://autoricardo-ag.slack.com/archives/C8RRP2UQ4/p1624284136079200

lkappeler (approved):

PR #317, opened 2021-07-21
DH: Show Whatsapp leads on lead center

nels64:
References CAR-8076

## Motivation and context

As a dealer, when checking my leads information on DH, I want to have a complete overview from leads including Whatsapp, so I can holistically understand the situation with my leads


src/services/car/messageLead.ts
```diff
@@ -245,3 +246,75 @@ export const resendMessageLead = async ({
     result: {},
   }
 }
+
+export const fetchDealerWhatsAppLeads = async ({
```
masone, line 250: this endpoint is on the reporting service, but the file is located in the `car/service/messageLead`, so it seems to be misplaced

nels64: I saw that we have all other endpoint there (e.g fetchDealerMessageLeads or fetchDealerCallLeads), do you want me to add add those endpoint into separate files?

masone: uhmm the convention would be to structure the methods / files / folders by the api structure. let's keep what we have now and clean up separately / globally

README.md
```diff
@@ -110,8 +110,10 @@ Also accompanying modes and param types, as well as default values, are exported
   - [`sendMessageLead`](https://carforyou-service.preprod.carforyou.ch/swagger-ui/index.html#/Lead/createMessageLeadUsingPOST)
   - [`fetchDealerMessageLeads`](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getAllDealerMessageLeadsUsingGET)
   - [`fetchDealerCallLeads`](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getAllDealerCallLeadsUsingGET)
+  - [`fetchDealerWhatsAppLeads`](https://reporting-service.preprod.carforyou.ch/swagger-ui/index.html#/Reporting/getAllDealerWhatsAppLeadsUsingGET)
```
talamcol, line 113: there is already an entry in the README for `postWhatsappTrackingEntry` could you please move it up there?

src/lib/factories/leads.ts
```diff
@@ -64,3 +86,10 @@ export const SearchCallLead = (attributes = {}): SearchCallLeadType => ({
   ...searchCallLeadDefaults,
   ...attributes,
 })
+
+export const SearchWhatsAppLead = (
+  attributes = {}
```
talamcol, line 91: I know you followed the pattern of the other factories but we could define `attributes` from being of the type `Partial<SearchWhatsAppLeadType>` with that we can support the consumer from passing the correct attributes. What do you think?

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -1,8 +1,10 @@
 import {
   fetchDealerCallLeads,
   fetchDealerMessageLeads,
+  fetchDealerWhatsAppLeads,
```
talamcol, line 4: In case you are planning to consider my comment in the README, the placement of these tests also needs to be adjusted

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -364,6 +367,80 @@ describe("Car API", () => {
...
+            size: 7,
+            sort: {},
+          },
+        })
+      } catch (err) {
+        error = err
+      }
+
+      expect(fetch).not.toHaveBeenCalled()
+      expect(error).toBeDefined()
```
talamcol, line 422: I'm not 100% sure here but I think you can move the two error asserts into the catch block then you can remove the let variable at the top

src/services/car/messageLead.ts
```diff
@@ -245,3 +246,75 @@ export const resendMessageLead = async ({
...
+      searchQuery && searchQuery.length >= 3
+        ? encodeURIComponent(searchQuery)
+        : "",
+  }
+
+  const path = `dealers/${dealerId}/whats-app-tracking-entries?${toQueryString(
+    queryParams
+  )}`
+
+  return await fetchPath({
```
talamcol, line 285: I think the `await` is not needed here, no?

src/types/models/index.ts
```diff
@@ -180,6 +180,16 @@ export interface SearchCallLead {
   duration: number
 }
 
+export interface SearchWhatsAppLead {
```
talamcol, line 183: the other whatsapp type/function is written as `Whatsapp` and not `WhatsApp` can we please be consistent and either adjust the old one or name the new one similar? Thanks!

README.md
nels64: not sure if it is a good idea, we have the same structure for `postCallTrackingEntry` as well.

src/services/car/__tests__/messageLead.test.ts
nels64: https://github.com/carforyou/carforyou-api-client-pkg/pull/317#discussion_r678076897

src/services/car/messageLead.ts
nels64: it needed since `fetchPath` is async function

src/services/car/__tests__/messageLead.test.ts
nels64: I tried but it doesn't work.

README.md
talamcol: yes but BE also groups them together and it's part of the reporting service. But up to you

src/services/car/__tests__/messageLead.test.ts
talamcol: checked out your branch. Works perfectly for me:
```javascript
it("should trow error if accessToken is not passed", async () => {
      try {
        await fetchDealerMessageLeads({
          dealerId: 1234,
          query: {
            page: 0,
            size: 7,
            sort: {},
            searchQuery: "",
          },
        })
        expect(fetch).not.toHaveBeenCalled()
      } catch (err) {
        expect(err).toBeDefined()
        expect(err.message).toBeDefined()
      }
    })
```

src/services/car/messageLead.ts
talamcol: I'm pretty sure you can safely do `return fetchPath()`. Because the promise is already awaited inside fetchPath. Also, all other functions except the messageLeads are returning it directly. Is it somehow different for your use case?

src/lib/factories/leads.ts
```diff
@@ -55,6 +56,27 @@ const searchCallLeadDefaults: SearchCallLeadType = {
...
+    image:
+      "2019/09/17/11/13/13/1-aixam-mac-500-cabriolet-480-261485-7e5hF5fhFUiC.jpg",
+    id: 555,
+    make: "Ford",
+    mileage: 6000,
+    model: "Fiesta",
+    price: 20000,
+    referenceId: "123",
+    type: "1.0 EcoB 125 Titanium X",
+  },
```
Averethel, line 75: Observation - we repeat that for every lead factory. Would it make sense to extract that to a factory of its own?

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -1,8 +1,11 @@
+/* eslint-disable jest/no-try-expect */
```
Averethel, line 1: Please not disable linter rules for the entire file, only do it for the relevant lines instead.

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -364,6 +368,78 @@ describe("Car API", () => {
...
+            size: 7,
+            sort: {},
+            searchQuery: "",
+          },
+        })
+        expect(fetch).not.toHaveBeenCalled()
+      } catch (err) {
+        expect(err).toBeDefined()
+        expect(err.message).toBeDefined()
+      }
```
Averethel, line 422: Oh, now I see why you need to disable the linter. Jest supports testing asynchronous code via helpers. In this case [`.rejects` helper](https://jestjs.io/docs/tutorial-async#rejects) should do the trick.

src/services/car/messageLead.ts
```diff
@@ -245,3 +246,75 @@ export const resendMessageLead = async ({
...
 }
+
+export const fetchDealerWhatsappLeads = async ({
+  dealerId,
+  query,
+  options = {},
+}: {
+  dealerId: number
+  query: LeadQueryParams
+  options?: ApiCallOptions & { validateOnly?: boolean }
```
Averethel, line 257:
```suggestion
  options?: ApiCallOptions
```

Why would you need `validateOnly` in the context of fetching a collection from the API?

src/services/car/messageLead.ts
```diff
@@ -245,3 +246,75 @@ export const resendMessageLead = async ({
...
+
+  const { sortOrder, sortType } = sort
+
+  const sortOrDefault = {
+    sortType: sortType || defaultLeadSort.sortType,
+    sortOrder: sortOrder || defaultLeadSort.sortOrder,
+  }
+
+  const queryParams = {
+    page: pageOrDefault(page, defaultDealerMessageLeadsPagination),
```
Averethel, line 270: Observation: `defaultDealerMessageLeadsPagination` is used in `call`, `message` and  now `whatsapp` leads. Let's rename it to `defaultLeadsPagination` to keep it in line with default options for sort

src/services/car/messageLead.ts
```diff
@@ -245,3 +246,75 @@ export const resendMessageLead = async ({
...
+    options: {
+      ...otherOptions,
+      isAuthorizedRequest: true,
+    },
+  })
+}
+
+export const hideWhatsappLead = async ({
+  dealerId,
+  whatsAppTrackingEntryId,
```
Averethel, line 296: I'd go with 
```suggestion
  id,
```
here. It's self explanatory from the context of the method name that it's the lead (or tracking entry) id

src/types/models/index.ts
```diff
@@ -180,6 +180,16 @@ export interface SearchCallLead {
   duration: number
 }
 
+export interface SearchWhatsappLead {
```
Averethel, line 183: Would it make sense to extract common parts of Message, Call and WhatsApp leads to a shared interface?

Averethel (changes requested):
A very solid starting point 👍🏻 

I have a few comments on the naming. Blocker for me is the testing and how the linter rule is disabled. Please take a look at that.

Migmar-lhamo:
@nels64 Can you please add the link of the Jira?

nels64:
> @nels64 Can you please add the link of the Jira?

https://autoricardo.atlassian.net/browse/CAR-8076

src/lib/factories/leads.ts
```diff
@@ -1,22 +1,25 @@
...
-    referenceId: "123",
-    image:
-      "2019/09/17/11/13/13/1-aixam-mac-500-cabriolet-480-261485-7e5hF5fhFUiC.jpg",
-    make: "Ford",
-    mileage: 6000,
-    model: "Fiesta",
-    price: 20000,
-    type: "1.0 EcoB 125 Titanium X",
-  },
+  listing: listing,
```
Averethel, line 22:
```suggestion
  listing,
```

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -364,6 +367,80 @@ describe("Car API", () => {
...
+          /\/dealers\/1234\/whats-app-tracking-entries\?page=1&size=7&sort=createdDate%2Cdesc$/
+        ),
+        expect.any(Object)
+      )
+    })
+
+    it("should trow error if accessToken is not passed", async () => {
+      let error
+      try {
+        await fetchDealerCallLeads({
```
Averethel, line 409: Heads up, you're calling another method that you promised testing. You probably want
```suggestion
        await fetchDealerWhatsappLeads({
```

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -364,6 +367,80 @@ describe("Car API", () => {
...
+            sort: {},
+          },
+        })
+      } catch (err) {
+        error = err
+      }
+
+      expect(fetch).not.toHaveBeenCalled()
+      expect(error).toBeDefined()
+      expect(error.message).toBeDefined()
```
Averethel, line 423: What's wrong with: 
```typescript
        expect(fetchDealerWhatsappLeads({
          dealerId: 1234,
          query: {
            page: 0,
            size: 7,
            sort: {},
          },
        })).rejects.toEqual("<error message>")
        expect(fetch).not.toHaveBeenCalled()
```
?

nels64: Expected: "[Error: You tried to make an authenticated requests without providing an access token! Please pass a valid token as a request option.]"
    Received: [Error: You tried to make an authenticated requests without providing an access token!
     Please pass a valid token as a request option.]

      414 |           },
      415 |         })
    > 416 |       ).rejects.toEqual(
          |                 ^
      417 |         "[Error: You tried to make an authenticated requests without providing an access token! Please pass a valid token as a request option.]"
      418 |       )
      419 |       expect(fetch).not.toHaveBeenCalled()


it works like this 
```
  await expect(fetchDealerWhatsappLeads({
          dealerId: 1234,
          query: {
            page: 0,
            size: 7,
            sort: {},
          },
        })).rejects.toBeDefined()
        expect(fetch).not.toHaveBeenCalled()
```

src/services/car/__tests__/messageLead.test.ts
```diff
@@ -364,6 +367,74 @@ describe("Car API", () => {
...
+      await expect(
+        fetchDealerWhatsappLeads({
+          dealerId: 1234,
+          query: {
+            page: 0,
+            size: 7,
+            sort: {},
+          },
+        })
+      ).rejects.toBeDefined()
```
Averethel, line 416: This worked for me:

```suggestion
      ).rejects.toEqual(
        new Error(
          "You tried to make an authenticated requests without providing an access token!\n Please pass a valid token as a request option."
        )
      )
```

Averethel (approved):

PR #318, opened 2021-07-22
feat: add charging data for e-cars

mreljin:
References https://autoricardo.atlassian.net/browse/CAR-7760

## Motivation and context

Adds the following charging data for e-cars:
- chargePort
- chargePower
- chargeSpeed
- fastChargePort
- fastChargePower
- fastChargeSpeed


src/lib/factories/listing.ts
```diff
@@ -71,6 +71,12 @@ const defaults: ListingType = {
   range: 153,
   rangeExtraUrban: 100,
   rangeUrban: 53,
+  chargePort: ["type-1"],
```
Averethel, line 74: are those values predefined and fetched from the backend? shouldn't this be added to reference data too?
The same goes for the fast charge port

Averethel (approved):
The code looks good, please verify the pre-fetching before merging.

src/lib/factories/listing.ts
mreljin: I may have made a mistake here, the line you reviewed should just be a simple string I think `"type-1"`, as when we're sending the form it sends a string. When prefetching it gets an array of charge ports: `["type-1", "type-2"]`
Correct me if I'm wrong

PR #337, opened 2021-09-02
feat: new method to fetch makes suggestions

talamcol:
References (hackdays)

## Motivation and context

for the full text search recommendations, backend introduced a new API

Bildschirmfoto 2021-09-07 um 14 04 42
(the part on the left)


Averethel (approved):

ufkaya (approved):
LGTM
Can you add a test?

talamcol:
> LGTM
> Can you add a test?

what do you want to test? since I'm neither doing something with the query nor with the response I'm not sure what I should test

ufkaya:
> > LGTM
> > Can you add a test?
> 
> what do you want to test? since I'm neither doing something with the query nor with the response I'm not sure what I should test

Usually, we try to add a test, even if it is a simple one, you can check `currentMakeModels.test.ts`, up to you

src/services/search/currentMakeModels.ts
```diff
@@ -19,3 +19,15 @@ export const fetchCurrentModels = async ({
...
 }
+
+export const fetchMakesSuggestions = async ({
+  query,
+  options = {},
+}: {
+  query: string
+  options?: ApiCallOptions
+}): Promise<MakeWithModels[]> =>
+  await fetchPath({
```
lkappeler, line 30: I guess you don't wanna await that rather return it?

talamcol: both works, will change it. Since the function is declared as async, you have to await it anyway

lkappeler: Oh, I missed that you used a shorthand arrow function on multiple lines...

PR #338, opened 2021-09-02
feat: return api response that contains the top facets for the popular filter search

lkappeler:
References https://github.com/carforyou/carforyou-listings-web/pull/3525

The top facets are returned under a new root key, adapts types and function accordingly


skomarica (approved):
Looks good to me, at least the new response format part. But better to have somebody from the FE for the rest :-)

lkappeler:
Semantic realease does not like my brnachname :(

PR #340, opened 2021-09-02
Popular filters

lkappeler:
References https://github.com/carforyou/carforyou-listings-web/pull/3525

The top facets are returned under a new root key, adapts types and function accordingly


src/types/facets.ts
```diff
@@ -3,5 +3,5 @@ export interface Facet {
 }
 
 export interface Facets {
-  [key: string]: Facet
+  [key: string]: Facet | Facets
 }
```
Averethel, line 7: Isn't what you want something along the lines of:

```typescript
export type Facets = { [key: string]: Facet } & { topFacets: Facets }
```

The advantage would be that you don't need to "guess" the type for keys other than `topFacets`

Averethel (approved):

src/services/search/facets.ts
```diff
@@ -13,16 +13,14 @@ export const fetchFacets = async ({
   fields?: string[]
   options?: ApiCallOptions
 } = {}): Promise<Facets> => {
-  const json = await postData({
+  return postData({
```
Averethel, line 16: BREAKING CHANGE

lkappeler: yeah like I realized for both the type and this that what I actually want is `  return { ...facets, topFacets }` to keep the IF and add the new property...

PR #343, opened 2021-09-08
feat: send saved search feedback

mreljin:
References https://autoricardo.atlassian.net/browse/CAR-7952

## Motivation and context

Add API functions for sending saved search feedback


src/services/userNotification.ts
```diff
@@ -90,3 +90,28 @@ export const fetchSavedSearch = async ({
...
+}: {
+  reason: string
+  key: string
+  options?: ApiCallOptions
+}): Promise<WithValidationError> => {
+  try {
+    await postData({
+      path: `saved-searches/key/${key}/unsubscribe-feedback`,
+      body: { reason },
+      options,
```
GutierrezGo, line 107: I'm wondering if we need the authentication here. Can you double check it? Thanks!

mreljin: Checked! We don't need it, we have a UUID in the path: `/key/${key}`

lkappeler (approved):
LGTM

GutierrezGo (approved):

PR #344, opened 2021-09-15
BREAKING CHANGE: remove needs assessment

talamcol:
References [CAR-8474](https://autoricardo.atlassian.net/browse/CAR-8474)

## Motivation and context

We remove needs assessment completely from the app. Therefore, the fetch method is no longer needed.


Averethel (approved):
You're faster creating PRs than I'm reviewing it :)

Looks good to me, take care to add a `feat` prefix to the commit message IIRC `BREAKING CHANGE` needs to be in the commit body, but won't trigger semantic release on its own.

PR #348, opened 2021-09-21
feat: add extendSavedSearch

Marine-Berthier:
References [CAR-8479](https://autoricardo.atlassian.net/browse/CAR-8479?atlOrigin=eyJpIjoiNjQyMDdhMTkxNTAxNGExY2IzMWFiY2M4NWFkZmM5NTIiLCJwIjoiaiJ9)

## Motivation and context

Add extendSavedSearch

### DIFF
--- a/src/services/userNotification.ts
+++ b/src/services/userNotification.ts
@@ -90,3 +90,28 @@ export const fetchSavedSearch = async ({
     path: `saved-searches/key/${key}`,
     options,
   })
+
+export const sendSavedSearchFeedback = async ({
+  reason,
+  key,
+  options = {},
+}: {
+  reason: string
+  key: string
+  options?: ApiCallOptions
+}): Promise<WithValidationError> => {
+  try {
+    await postData({
+      path: `saved-searches/key/${key}/unsubscribe-feedback`,
+      body: { reason },
+      options,
+    })
+
+    return {
+      tag: "success",
+      result: {},
+    }
+  } catch (error) {
+    return handleValidationError(error)
+  }
+}
--- a/src/services/userNotification.ts
+++ b/src/services/userNotification.ts
@@ -79,6 +79,29 @@ export const enableSavedSearch = async ({
   }
 }
 
+export const extendSavedSearch = async ({
+  key,
+  options = {},
+}: {
+  key: string
+  options?: ApiCallOptions
+}): Promise<WithValidationError> => {
+  try {
+    await postData({
+      path: `saved-searches/key/${key}/extend`,
+      body: {},
+      options,
+    })
+
+    return {
+      tag: "success",
+      result: {},
+    }
+  } catch (error) {
+    return handleValidationError(error, { swallowErrors: true })
+  }
+}
+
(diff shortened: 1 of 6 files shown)


src/services/userNotification.ts
```diff
@@ -79,6 +79,29 @@ export const enableSavedSearch = async ({
...
+      body: {},
+      options,
+    })
+
+    return {
+      tag: "success",
+      result: {},
+    }
+  } catch (error) {
+    return handleValidationError(error, { swallowErrors: true })
```
lkappeler, line 101: What kind of validation error is expected to be returned from the endpoint?

src/types/models/index.ts
```diff
@@ -137,6 +137,7 @@ export interface SavedSearch {
   language: string
   searchQuery: ListingFilterParams
   uiMetadata: Record<string, unknown>
+  enabledUntil?: string
```
lkappeler, line 140: Will there be saved searches not expiring?

Marine-Berthier: I fixed that ! Thank you Lorenzo !

src/services/userNotification.ts
Marine-Berthier: It will return a 404

masone: I'm confused now. a validation error should not be a 404 not found. did you mean 400 bad request? I think Lorenzo is asking what validations this endpoint does on which fields. maybe worth that you two are getting in a quick call to clear up the question/answer

masone: (with the 404 you might have intended to say that if the key is wrong, it would be a 404. in that case, there would be no need to handle validation errors because there are no actual fields that get validated. the 404 should be properly raised and handled by the app imo)

Marine-Berthier: If the key is wrong it will return a 404. The same happens for `enableSavedSearch`, we have a 404 if the key is wrong but we still handle validation errors in this case, so I am confused 😕

src/types/models/index.ts
Marine-Berthier: Actually, I think it should remain optional. When the savedSearch is sent to BE the first time, we don't want to specify the expiration date since this data will be generated by BE. I noticed this when I tried this new version on my branch, I have a type error stating that `enabledUntil` is missing[ here ](https://github.com/carforyou/carforyou-listings-web/blob/678b653ea9ff023ace2ba6e73715f69d1ac21f5b/components/forms/savedSearchForm.tsx#L57)

lkappeler (approved):
LGTM

src/types/models/index.ts
lkappeler: We discussed in a call: it is only not present when creating a saved search so we decided to keep it mandatory as part of the type and omit it for creation.

src/services/userNotification.ts
lkappeler: Marine and I discussed that it is not needed to add validation error handling in this case as the only error response sent is a 404 for an invalid key, in comparison to forms where we receive detailed information about the post body validation

PR #350, opened 2021-09-30
fix: add a fallback value for gbdScore

masone:
References https://autoricardo.atlassian.net/browse/CAR-8584

## Motivation and context

some listings on the search service can have `null` values for `gdbdScore`, but our code expects a valid value from an enum


masone:
Already on it, wanted a quick verification that I'm in the right place ;)

Averethel (approved):

PR #351, opened 2021-10-01
feat: add API to delete user

ufkaya:
References [CAR-8470](https://autoricardo.atlassian.net/browse/CAR-8470)

## Motivation and context

Users and Dealers can delete their account


src/services/user.ts
```diff
@@ -0,0 +1,22 @@
...
+      path: `users/me`,
+      options: { isAuthorizedRequest: true, ...options },
+    })
+
+    return {
+      tag: "success",
+      result: {},
+    }
+  } catch (error) {
+    return handleValidationError(error)
```
lkappeler, line 20: What kind of validation error are we getting from a delete endpoint?

src/services/user.ts
```diff
@@ -0,0 +1,22 @@
+import { WithValidationError } from "../types/withValidationError"
```
lkappeler, line 1: As [mentioned](https://autoricardo-ag.slack.com/archives/C0289PXUV8A/p1633096739000300?thread_ts=1632823022.003600&cid=C0289PXUV8A) by Slavko, the User endpoints are currently part of the dealer service, what's the thought behind putting them into a separate file and not the one handling the endpoints for the dealer service?

src/services/user.ts
ufkaya: oh..actually we don't, so will re-add swallowErrors again

src/services/user.ts
ufkaya: I missed that and since it is under Users I created a new file

src/services/user.ts
lkappeler: I meant more like the handleValidationError function, in general, is very specific to how we handle form submissions that are having backend validations. In this case I would backend expect to only return status codes like 200/401 which are the only two cases we will ever have on deleting a user....

ufkaya: I would still keep it for consistency, this is what we do in other places

lkappeler (approved):

src/services/user.ts
lkappeler: As described, I think the majority of times we do that is to handle form validation. I don't see a point in adding envelopes designed to handle and display validation errors if there is no case.

src/services/dealer.ts
```diff
@@ -318,3 +319,23 @@ export const putUserAccount = async ({
     return handleValidationError(error, { swallowErrors: true })
   }
 }
+
+export const deleteUser = async ({
```
masone, line 323: can you move this to the user service please? on our side, it's currently on the dealer service, which is not only off with backends structure, but also with our own now that we have non-dealer users

https://dealer-service.preprod.carforyou.ch/swagger-ui/index.html#/User/deleteUserUsingDELETE

ufkaya: it will be moved to the dealer service https://github.com/carforyou/carforyou-api-client-pkg/pull/351#discussion_r722128813

PR #354, opened 2021-10-01
CAR-8706 update account

lkappeler:
References https://autoricardo.atlassian.net/browse/CAR-8387

## Motivation and context

Extend the API client with an endpoint to update user account data

### DIFF
--- a/src/types/models/account.ts
+++ b/src/types/models/account.ts
@@ -4,7 +4,7 @@ export interface UserAccount {
   firstName: string
   lastName: string
   address: string
-  zipCode: number
+  zipCode: string
   city: string
   phone: string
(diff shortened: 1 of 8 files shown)


src/types/models/account.ts
```diff
@@ -0,0 +1,11 @@
+export type Salutation = "mr" | "mrs" | "other"
+
+export interface UserAccount {
+  firstName: string
+  lastName: string
+  address: string
+  zipCode: number
+  city: string
+  phone: string
+  salutation: Salutation
```
GutierrezGo, line 10: Maybe you can pick a set of selected keys from `BuyNowApplication`.

`Pick<Type, Keys>`

PR #379, opened 2021-11-08
feat: add inspectedEffective as part of ListingBasicData type

gzimh:
References [CAR-9039](https://autoricardo.atlassian.net/browse/CAR-9039)

## Motivation and context
Adds prop `inspectedEffective` to `ListingBasicData` type.


masone (approved):

lkappeler:
@gzimh / @masone What does `inspectedEffective` mean?

masone:
The original ticket is here https://autoricardo.atlassian.net/browse/CAR-6844
I understood that we have multiple sources leading to the fact that a car is marked "ab MFK". One is the `inspected` field, which we get from the dealer through the imports. But there are also cases where we calculate the value based on other information. `inspectedEffective` is the calculated field, which should be displayed to the user. Backend considered keeping the exposed field stable, but ended up introducing a new one for the sake of reducing mappings on their end

PR #380, opened 2021-11-09
BREAKING CHANGE: remove fields from facets api

talamcol:
References [CAR-9027](https://autoricardo.atlassian.net/browse/CAR-9027)

## Motivation and context

Backend deprecated the fields property and we should use `facets` now

* listings is already cleaned
* dealerhub PR is [here](https://github.com/carforyou/carforyou-dealerhub-web/pull/1653)
* other projects are not affected


masone (approved):
because we missed it a couple of times in the last weeks - make sure to have the BREAKING CHANGE in the commit body. you'll still need some commit title:

```
chore: barfoo

BREAKING CHANGE: foobar
```

lkappeler (approved):

PR #381, opened 2021-11-09
feat: new methods for loan interest

talamcol:
References [CAR-9003](https://autoricardo.atlassian.net/browse/CAR-9003)

## Motivation and context

[Describe the problem we're solving and lay out the solution]


README.md
```diff
@@ -256,6 +258,8 @@ Also accompanying modes and param types, as well as default values, are exported
   - [`sendBuyNowApplication`](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Buy%20Now%20Application)
   - [`fetchBuyNowConfiguration`](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Buy%20Now%20Configuration)
   - [`markBuyNowApplicationAsPaid`](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Buy%20Now%20Application/markAsPaidUsingPOST)
+- [Loan](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Loan%20Interest)
+  - [`createLoanInterest`](https://buyer-service.preprod.carforyou.ch/swagger-ui/index.html#/Loan%20Interest/createUsingPOST)
```
Averethel, line 262: Don't we need docs for `calculateMonthlyRate` too?

src/index.ts
```diff
@@ -31,10 +31,15 @@ export {
   DealerAnalytics,
   BuyNowConfiguration,
   MessageLeadListing,
+  EmploymentType,
+  Gender,
```
Averethel, line 35: Do we ask for gender in any of the forms in listings? IIRC it was used in MBG, but that's gone, but online purchase or buyer protection? If yes let's align listings and use exposed enum there.

src/types/models/listing.ts
```diff
@@ -132,6 +132,7 @@ export interface Listing
   interior: string
   kiloWatts: number
   leasingMonthlyRate: number
+  loanMonthlyRate: number
```
Averethel, line 135: Shouldn't this be optional?

src/types/models/listing.ts
```diff
@@ -201,6 +202,7 @@ interface BaseSearchListing {
   modelKey: string
   price: number
   leasingMonthlyRate: number
+  loanMonthlyRate: number
```
Averethel, line 205: Shouldn't this be optional?

Averethel (approved):
Looks good to me. Please do a sanity check on my questions before merging.

src/types/models/listing.ts
mreljin: BE sends both `leasingMonthlyRate` and `loanMonthlyRate`. if `leasingMonthlyRate` exists -> the user is ineligible for a loan and `loanMonthlyRate` is null. And the other way around. Confirmed with BE

src/index.ts
mreljin: > but online purchase or buyer protection?

Checked, neither use gender in any form

PR #398, opened 2021-11-24
feat(fetchMessageLeads): fetch message leads for the current user

lkappeler:
References https://autoricardo.atlassian.net/browse/CAR-9204

## Motivation and context

Extends the API client with a new endpoint to fetch the message leads for the current user


src/lib/factories/reporting.ts
```diff
@@ -0,0 +1,30 @@
...
+    model: "Mustang",
+    price: 30000,
+    referenceId: "ABCFEDG",
+    type: "GT",
+  },
+}
+
+export const userMessageLeadFactory = (
+  props: Partial<
+    Pick<UserMessageLead, "createdDate" | "id"> & {
```
GutierrezGo, line 20: 😍

src/services/reporting/__tests__/messageLeads.test.ts
```diff
@@ -0,0 +1,16 @@
+import { fetchMessageLeads } from "../messageLeads"
+
+describe("Reporting Service", () => {
+  describe("fetchMessageLeads", () => {
+    it("calls the api", async () => {
+      await fetchMessageLeads({
+        options: { accessToken: "Let me in, please." },
```
GutierrezGo, line 7: 🤭

src/types/models/reporting.ts
```diff
@@ -0,0 +1,18 @@
+import { Listing } from "../../index"
+
+export type UserMessageLeadListing = Pick<
```
GutierrezGo, line 3: 😍

src/services/reporting/messageLeads.ts
```diff
@@ -0,0 +1,21 @@
...
+  options = {},
+}: {
+  offset?: number
+  page?: number
+  size?: number
+  options?: ApiCallOptions
+} = {}): Promise<Paginated<UserMessageLead>> => {
+  const query = toQueryString({ offset, page, size })
+  return fetchPath({
+    path: `users/me/message-leads${query ? "?" + query : ""}`,
```
GutierrezGo, line 18:
```suggestion
    path: `users/me/message-leads${query ? `?${query}` : ""}`,
```

GutierrezGo (approved):
It looks good to me.
Thanks 🧸

src/services/reporting/messageLeads.ts
masone: what happens when `query` is an empty object? :)

src/lib/factories/__tests__/reporting.test.ts
```diff
@@ -0,0 +1,24 @@
+import { defaultUserMessageLead, userMessageLeadFactory } from "../reporting"
+
+describe("Reporting service factories", () => {
```
Averethel, line 3: Isn't testing the factories a bit of an overkill?

src/lib/factories/reporting.ts
```diff
@@ -0,0 +1,30 @@
...
+      listing: Partial<UserMessageLeadListing>
+    }
+  > = { listing: {} }
+): UserMessageLead => {
+  return {
+    ...defaultUserMessageLead,
+    ...props,
+    listing: { ...defaultUserMessageLead.listing, ...props.listing },
```
Averethel, line 28: Maybe we can leverage the listing factory here?

src/services/reporting/__tests__/messageLeads.test.ts
Averethel: giphy (7)

src/services/reporting/messageLeads.ts
Averethel: Query is always a string so if it's empty nothing will be appended to the URL

masone: oh, it runs through `toQueryString` above - thanks! I thought to have identified it as an object in the types, but I probably didn't check properly and assumed too much

masone (approved):

src/lib/factories/reporting.ts
lkappeler: The problem with that is the response only contains a subset of mapped listing data so it would not match the type or like have too much data this may actually lead to some problems when we expect data that is not returned on runtime.

src/lib/factories/__tests__/reporting.test.ts
lkappeler: Hmm maybe, it does not hurt I guess. (I would have actually forgotten to spread the listings object without the test...)

PR #404, opened 2021-11-30
BREAKING CHANGE: Replace make suggestions method

talamcol:
References [CAR-8727](https://autoricardo.atlassian.net/browse/CAR-8727)

Backend implemented a more efficient way to recommend make/model types and deprecated the old API.


GutierrezGo (approved):
Sorry, is it possible that you update the `lockfileVersion` in your PR?
As far as I know, the first person that touch the pkg could update it.  (I know you are not the first one BUT just if it doesn't bother you to do it in your lovely PR).

Thanks 🥦

Averethel (approved):

PR #414, opened 2021-12-13
BREAKING CHANGE: Remove body color group & add body color 

ShpatHyseni:
References [CAR-9061](https://autoricardo.atlassian.net/browse/CAR-9061)

## Motivation and context

Add  `bodycolor` to `ListingFilterParams` type.


src/types/models/listing.ts
```diff
@@ -113,6 +113,7 @@ export interface Listing
   typeSlug: string
   active: boolean
   bodyColorGroup: string
```
ufkaya, line 115: Do we still need it?

ShpatHyseni: I think not, but I left it just in case some project is still using bodyColorGroup prop.
We can remove it later.

ufkaya: removing is already part of the task, please check if you still need it, and if not, let's remove it and release a breaking change

masone: plus one. backend will likely remove it on their end anyway

ufkaya (approved):
👍

masone (approved):
thanks, looking good! please validate the format of the commit message before merging. you might be aware, but just to be sure - the one in the title won't trigger a release as-is. the `BREAKING CHANGE` indicator belongs to the commit message body, while it still needs a usual "trigger" in the subject, eg:

```
fix: replace bodyColorGroup with bodyColor

BREAKING CHANGE: bodyColorGroup is removed
```

PR #426, opened 2022-01-04
feat: add favourites apis

Averethel:
References CAR-9341

## Motivation and context

Adds APIs to support the management of user favourites


src/services/favourites.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const saveFavourite = async ({
+  listingId,
+  options = {},
+}: {
+  listingId: number
+  options?: ApiCallOptions
+}): Promise<Response> => {
+  return postData({
```
talamcol, line 36: just to confirm, is there no validation error for this one if for example the listing is no longer active or something like that?

talamcol (approved):
looks good to me!

src/services/favourites.ts
Averethel: Technically there is but the way we use the APIs we guarantee that the listing exists (since you favourite it from the SRP or PDP where it's displayed)

PR #427, opened 2022-01-06
feat: build with tsc

masone:
References https://autoricardo.atlassian.net/browse/CAR-9284

## Motivation and context

Moving away from pika/pack, following up on the RFC in https://github.com/carforyou/carforyou-docs/discussions/147 - this POC creates an esm and commonjs build with esm. similar result to what we had before, just without the bloat of pika/pack


masone:
this has proven the point for the POC, but I'll start over fresh

PR #430, opened 2022-01-11
fix: building package with tsc

masone:
References carforyou/carforyou-docs#147

## Motivation and context

- [x] tested prerelease

Moving away from pika/pack, following the decisions outlined in the RFC above. Zero dependencies for a build

- removed pika/pack
- removed bundlesize check, see https://autoricardo-ag.slack.com/archives/C018LPJG9K7/p1641903077011500
- moved `__tests__` in root to `src/__tests__` to be covered by tsconfig


tsconfig.json
```diff
@@ -1,13 +1,13 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
-    "target": "es5",
-    "lib": ["es5", "es6", "es7", "es2017", "dom"],
-    "rootDirs": ["src"],
-    "baseUrl": "src",
```
Averethel, line 7: Can we keep those? This unlocks not pre-pending loads of `../` to imports

Averethel (approved):

tsconfig.json
masone: dang, I missed your comment. we have been using them in one case only. I'm happy if we reintroduce / harmonize across packages at some other point in time

PR #439, opened 2022-01-21
feat: align embedded listing type in various responses

Averethel:
References CAR-9593

Introduces a SimpleSearchListing type that had been duplicated in
- buyer feedback batch
- car sales
- call leads
- whatsapp tracking entries
- message leads
- questions
- favourites

BREAKING CHANGE: image fields in those type was deprecated and replaces with an array of image objects including s3Key (similar to SearchListing and Listing) for consistency
This allows us to remove bunch of manual re-mapping done on the FE side


lkappeler (approved):
The changes are looking fine to me, the naming is not super speaking to me as "Simple" does not really give me an interaction on how it differs from a regular search listing. Perhaps something like `PartialSearchListing` or `MinimalSearchListing` instead? (Doe we know if/how backend calls them)

PR #461, opened 2022-02-14
feat: add new methods to bulk fetch data for comparison tool

talamcol:
References [CAR-9729](https://autoricardo.atlassian.net/browse/CAR-9729)

## Motivation and context

To fetch the data for the comparison tool, bulk fetch methods have been introduced. To see how they are used, here is the related PR in listings: https://github.com/carforyou/carforyou-listings-web/pull/4137

### DIFF
--- a/src/services/car/inventory.ts
+++ b/src/services/car/inventory.ts
@@ -1,7 +1,7 @@
-import { BulkFetchResponse } from "../../types/models"
 import { WithValidationError } from "../../types/withValidationError"
 import { Paginated } from "../../types/pagination"
 import { Listing } from "../../types/models/listing"
+import { BulkFetchResponse } from "../../types/models"
 import toQueryString from "../../lib/toQueryString"
 import { decodeDate, encodeDate } from "../../lib/dateEncoding"
 import {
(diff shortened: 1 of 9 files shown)


src/services/car/inventory.ts
```diff
@@ -41,6 +42,25 @@ export const fetchListing = async ({
...
+}): Promise<BulkFetchResponse<Listing>[]> => {
+  const listings = await postData({
+    path: "listings/bulk-get",
+    body: { elements: listingIds },
+    options,
+  })
+
+  return listings.map((listing) => ({
+    id: listing.id,
+    payload: sanitizeListing(listing.payload),
```
lkappeler, line 60: There seems to be some specific logic regarding date handling not quite sure what the intention behind feels a bit as it would be for a specific use case at first glance...
I would either drop it or add a test to make sure it's returned as expected

lkappeler (approved):
Looks fine to me, can you check the comments on the dates before you merge it.

src/services/car/inventory.ts
talamcol: the method should return the same as `fetchListing` but in bulk and not just a single one. That's why I added it. `sanitizeListing` does some magic with the dates.

talamcol: added a test for it

PR #463, opened 2022-02-16
feat: fix api aggregations

ufkaya:
References CAR-XXXX

## Motivation and context

[Describe the problem we're solving and lay out the solution]

### DIFF
--- a/src/services/search/listingSearch.ts
+++ b/src/services/search/listingSearch.ts
@@ -284,8 +284,8 @@ export const fetchAggregations = async ({
   const response = await postData({
     path: "listings/aggregations",
     body: {
-      query,
-      aggregations: [{ name: "locationSimple" }],
+      query: paramsToSearchRequest(query),
+      aggregations: [{ name: "LOCATION" }],
     },
     options,
   })


ufkaya:
> I am a bit confused why we need the change in the API client / for the A/B test?
> 
> As discussed in the last refinement and mentioned by Slvko in Slack we have V1 for the A/B test https://autoricardo-ag.slack.com/archives/C02T9LUQ806/p1644427783893509 It'susing `locationSimple` aggregations to provide the location buckets.
> 
> And then we have the V2 for the spike where we explore performance and adaptations for a server-side aggregation So I would not expect V2 to be consumed for the A/B test as it may change depending on the outcomes of the spike.
> 
> > API is production ready, but it can be adapted, depending on the outcome of the spike: https://autoricardo.atlassian.net/browse/CAR-9715

Might be that I am missing something, I dont know what happened, I cant use the V1, there is only 1 API https://internal.carforyou.dev/api-docs/swagger-ui/?urls.primaryName=inventory-search-service#/Listing%20Search/getAggregations which is the V2

ufkaya:
> Perhaps he missed merging the docs when deploying the V2, V1 works perfectly fine when I test it in CURL
> 
> ```
> curl --location --request POST 'https://api.carforyou.ch/listings/aggregations' \
> --header 'Content-Type: application/vnd.carforyou.v1+json' \
> --data-raw '{
>     "query": {
>         "bodyType": ["suv"],
>         "priceFrom": 20000,
>         "priceTo": 30000
>     },
>     "aggregations": [
>         {
>             "name": "locationSimple"
>         }
>     ]
> }'
> ```
> 
> The PoC for the performance also seems to work fine, maybe you can check if something in the request is different from the code... At first glance, it might by query formatting `query: paramsToSearchRequest(query)` from the div that causes some issues in V1.
> 
> API request: https://github.com/carforyou/carforyou-listings-web/pull/4093/files#diff-22a84923125adf610906ea48c48232e48c1f5b48eacd3cca7af7c8bb38bf5608R188
> 
> UI: https://lkappeler-playground-map-stugg.carforyou-listings-web.branch.carforyou.ch/en/auto/search/map?transmissionType=automatic&fuelTypeGroup=petrol

yep V1 works too, thanks

lkappeler (approved):

PR #465, opened 2022-02-21
feat: add similarTo to ListingQueryParams

ufkaya:
References [CAR-9676](https://autoricardo.atlassian.net/browse/CAR-9676)

## Motivation and context

> Due to the expected increased amount of traffic, on the backend we have optimized the search API.  BE added similarTo in ListingSearch query, FE should change the logic and instead of similarToListingId fill similarTo with the appropriate listing values and fill in the idNot field with the id of that listing.


lkappeler (approved):
Looking good, maybe add a test case to cover for the new params?

PR #468, opened 2022-02-25
fix: date Type

Marine-Berthier:
Fix Date Interface. [Related PR](https://github.com/carforyou/carforyou-listings-web/pull/4176)


Averethel (approved):
Do we always have the month and the year? I vaguely recall that month might be optional for some dates during manual listing creation. Could you verify that?

Marine-Berthier:
@Averethel Date is the type of `lastServiceDate`, `lastInspectionDate` and `firstRegistrationDate` and for all 3 it works as follow: 
- if only the year is filled, the month is set up to January by default
- if only the month is filled, then we considered that there's no date at all and we return `null`
It's either this format `{ year: number, month: number }`, or `null`
I tested it in both DH and listings and it worked fine but I'll ask QA to have a look before merging

talamcol (approved):
please don't forget to change the PR name before merge :D

PR #473, opened 2022-03-02
fix: remove systemPerformanceKiloWatts

Marine-Berthier:
References [CAR-9744](https://autoricardo.atlassian.net/browse/CAR-9744)

## Motivation and context

Remove `systemPerformanceKiloWatts`, this field was not used


ufkaya (approved):
LGTM 👍 
It should be a BREAKING CHANGE

